PR #82, opened 2026-02-10
fix: return 200 with ConnectionTestResult when pg is not installed

### LINKED ISSUE #80: POST /api/database/test returns 500 when pg is not installed instead of 200 with error body
## Bug

`POST /api/database/test` returns HTTP 500 when the `pg` (PostgreSQL client) module is not installed. The endpoint's contract is to always return 200 with success/failure in the JSON body, but the missing-`pg` case breaks this by calling `errorResponse(res, ..., 500)`.

This causes 3 e2e test failures in `test/database-api.e2e.test.ts`:

- `returns failure for unreachable host` — expects 200, gets 500
- `includes timing information` — expects `durationMs` in body, gets `{error: "..."}`
- `has correct response shape on failure` — expects `{success, serverVersion, error, durationMs}`, gets `{error}`

## Root Cause

In `src/api/database.ts`, `handleTestConnection()` dynamically imports `pg` at line 534. When the import fails (because `@elizaos/plugin-sql` isn't installed), the catch block at line 537 calls:

```typescript
errorResponse(res, "PostgreSQL client library (pg) is not available...", 500);
```

This returns `{error: "..."}` with status 500, bypassing the `ConnectionTestResult` shape (`{success, serverVersion, error, durationMs}`) that all callers expect.

## Proposed Fix

When `pg` is not available, return a proper `ConnectionTestResult` with status 200:

```typescript
jsonResponse(res, {
  success: false,
  serverVersion: null,
  error: "PostgreSQL client library (pg) is not available. Ensure @elizaos/plugin-sql is installed.",
  durationMs: Date.now() - start,
} satisfies ConnectionTestResult);
```

This preserves the endpoint's contract (always 200, result in body) and gives the caller a clean error message.

## Environment

- macOS, Node 23.10
- milaidy v2.0.0-alpha.9
- `pg` not installed (no `@elizaos/plugin-sql`)

wbaxterh:
## Summary

- `POST /api/database/test` was returning HTTP 500 when `pg` is not installed, breaking the endpoint's contract of always returning 200 with success/failure in the body
- Changed the missing-`pg` catch block from `errorResponse(res, ..., 500)` to `jsonResponse(res, { success: false, serverVersion: null, error: "...", durationMs })` matching the `ConnectionTestResult` type
- Fixes 3 e2e test failures in `test/database-api.e2e.test.ts`

## Test plan

- [x] `pnpm vitest run test/database-api.e2e.test.ts --config vitest.e2e.config.ts` — 36/36 passing
- [x] `pnpm vitest run --config vitest.e2e.config.ts` — 9 files passed, 327 tests passing
- [x] `pnpm vitest run --config vitest.unit.config.ts` — 42 files, 1033/1033 passing

Fixes #80

### DIFF
--- a/src/api/database.ts
+++ b/src/api/database.ts
@@ -534,11 +534,13 @@ async function handleTestConnection(
     const pgModule = await import("pg");
     Pool = pgModule.default?.Pool ?? pgModule.Pool;
   } catch {
-    errorResponse(
-      res,
-      "PostgreSQL client library (pg) is not available. Ensure @elizaos/plugin-sql is installed.",
-      500,
-    );
+    jsonResponse(res, {
+      success: false,
+      serverVersion: null,
+      error:
+        "PostgreSQL client library (pg) is not available. Ensure @elizaos/plugin-sql is installed.",
+      durationMs: Date.now() - start,
+    } satisfies ConnectionTestResult);
     return;
   }
 


lalalune:
Included in #84 — merged with formatting fixes.

PR #83, opened 2026-02-10
fix: harden cloud login browser open

lawyered0:
**Summary**
- Prevent shell injection when opening Eliza Cloud login URL.
- Remove shell wrapper from self-updater spawn.

**Changes**
- Validate URL protocol and open via execFile; handle missing opener binary.
- Remove shell:true from self-updater (apt path already uses sh -c explicitly).

**Testing**
- Not run (not requested).

**Risk**
- Low: change limited to browser-open flow and update process spawn.


lalalune:
Included in #84 — merged with formatting fixes.

PR #85, opened 2026-02-10
milaidy: harden plugins open editor

lawyered0:
**Summary**

Avoid shell execution when opening plugin directories via EDITOR.

**Changes**

Parse EDITOR into argv and run with spawnSync (no shell).

**Testing**

Not run (not requested).


lalalune (approved):
LGTM. Good security hardening - replaces execSync shell injection surface with spawnSync.

PR #105, opened 2026-02-10
fix: misc bugfixes (#103, #80)

### LINKED ISSUE #80: POST /api/database/test returns 500 when pg is not installed instead of 200 with error body

### LINKED ISSUE #103: Debug console.log and duplicate db adapter init in eliza.ts
## Bug

Two issues in `src/runtime/eliza.ts` from the recent merge-security-fixes PR:

### 1. Debug `console.log` left in production code (line 1671)

```typescript
if (sqlPlugin) {
  await runtime.registerPlugin(sqlPlugin.plugin);
  console.log("sqlPlugin", sqlPlugin);  // ← debug log
```

This prints plugin internals to stdout on every agent startup.

### 2. Duplicate database adapter initialization (lines 1696-1706)

The "7c" block that eagerly initializes the database adapter appears twice — once inside the `if (sqlPlugin)` branch (lines 1673-1683) and again unconditionally after it (lines 1696-1706). The second copy is a verbatim duplicate and should be removed.

```typescript
// First occurrence (line 1673) — correct, inside if (sqlPlugin)
if (runtime.adapter && !(await runtime.adapter.isReady())) {
  await runtime.adapter.init();
  logger.info("[milaidy] Database adapter initialized early (before plugin inits)");
}

// ... else branch throws ...

// Second occurrence (line 1696) — duplicate, should be removed
if (runtime.adapter && !(await runtime.adapter.isReady())) {
  await runtime.adapter.init();
  logger.info("[milaidy] Database adapter initialized early (before plugin inits)");
}
```

The call is idempotent (checks `isReady()`) so this doesn't cause a crash, but it's dead code and adds unnecessary async overhead on every startup.

## Proposed Fix

1. Remove the `console.log` on line 1671
2. Remove the duplicate block on lines 1696-1706

0xSolace:
## Summary
- remove stray sql plugin debug `console.log` from runtime startup
- remove duplicated early database adapter initialization block
- ensure `/api/database/test` returns a contract-safe `{ success: false, ... }` payload if PostgreSQL client initialization fails

## Issues
- Closes #103
- Closes #80

## Validation
- `npm run typecheck`

### DIFF
--- a/src/api/database.ts
+++ b/src/api/database.ts
@@ -576,12 +576,26 @@ async function handleTestConnection(
     return;
   }
 
-  const pool = new Pool({
-    connectionString,
-    max: 1,
-    connectionTimeoutMillis: 10000,
-    idleTimeoutMillis: 5000,
-  });
+  let pool: import("pg").Pool | null = null;
+  try {
+    pool = new Pool({
+      connectionString,
+      max: 1,
+      connectionTimeoutMillis: 10000,
+      idleTimeoutMillis: 5000,
+    });
+  } catch (err) {
+    jsonResponse(res, {
+      success: false,
+      serverVersion: null,
+      error:
+        err instanceof Error
+          ? err.message
+          : "Failed to initialize PostgreSQL client.",
+      durationMs: Date.now() - start,
+    } satisfies ConnectionTestResult);
+    return;
+  }
 
   let client: import("pg").PoolClient | null = null;
   try {
@@ -607,7 +621,7 @@ async function handleTestConnection(
     } satisfies ConnectionTestResult);
   } finally {
     if (client) client.release();
-    await pool.end();
+    if (pool) await pool.end();
   }
 }
 
--- a/src/runtime/eliza.ts
+++ b/src/runtime/eliza.ts
@@ -1881,7 +1881,6 @@ export async function startEliza(
   //     this.adapter is undefined, so plugins that use runtime.db will fail.
   if (sqlPlugin) {
     await runtime.registerPlugin(sqlPlugin.plugin);
-    console.log("sqlPlugin", sqlPlugin);
 
     // 7c. Eagerly initialize the database adapter so it's fully ready (connection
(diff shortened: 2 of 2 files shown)


lalalune:
Superseded by PR #106 which was merged. The debug log removal and duplicate adapter init fix are already in.

PR #139, opened 2026-02-11
milaidy: cap agent import decompression size

lawyered0:
## Impact
Prevents a memory-exhaustion denial-of-service during agent import by rejecting compressed payloads that expand beyond a fixed safe limit.

## Exploit path
`importAgent()` previously used unbounded `gunzipSync` on attacker-controlled encrypted payload content. A crafted gzip bomb could expand massively and consume process memory during import.

## Fix
- Replace unbounded sync gunzip with bounded streaming decompression in `src/services/agent-export.ts`.
- Enforce a hard decompressed-size cap (`16 MiB`) and throw deterministic `AgentExportError` once exceeded.
- Preserve normal import behavior for valid payloads.

## Tests
- Added regression test in `src/services/agent-export.test.ts`:
  - `rejects compressed payloads that expand beyond import limit`
- Ran:
  - `bunx vitest run src/services/agent-export.test.ts`
  - `bunx biome check src/services/agent-export.ts src/services/agent-export.test.ts`

## No-overlap statement
This PR addresses only agent import decompression bounds in the export/import service and does not overlap with prior non-loopback API auth or plugin/MCP key-validation fixes.

### DIFF
--- a/src/services/agent-export.test.ts
+++ b/src/services/agent-export.test.ts
@@ -860,6 +860,48 @@ describe("agent-export", () => {
       expect(result.success).toBe(true);
       expect(result.counts.memories).toBe(500);
     });
+
+    it("rejects compressed payloads that expand beyond import limit", async () => {
+      const { gzipSync } = await import("node:zlib");
+
+      // Highly compressible payload that expands to >16 MiB when decompressed.
+      const oversizedJson = JSON.stringify({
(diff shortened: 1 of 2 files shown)


lawyered0:
Closing and recreating PR to reset discussion metadata with sanitized content only.

PR #231, opened 2026-02-14
fix: review pipeline — comment posting and external PR support

0xSolace:
## Summary
- fix missing PR comments by requiring structured JSON output from claude-code-action and posting/updating a deterministic sticky PR comment via github-script
- derive merge/close verdict directly from the structured output path instead of polling for bot comments that may never be posted in automation mode
- allow external contributors to be reviewed on pull_request_target by enabling claude-code-action's `allowed_non_write_users` bypass (with existing workflow-scoped token permissions)

## Why this fixes it
- claude-code-action with a `prompt` runs in automation/agent mode, which does not create tracking comments by default; this workflow now posts the review comment explicitly
- external PR authors often have read-only permission and were blocked by the action's write-permission gate; `allowed_non_write_users` removes that blocker for review-only automation

## Notes
- keeps existing approval/reject/close automation behavior
- includes sticky marker `<!-- agent-review:claude-structured -->` to avoid comment spam

Co-authored-by: Sol <sol@shad0w.xyz>

### DIFF
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -86,7 +86,10 @@ jobs:
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          claude_args: '--model claude-opus-4-6'
+          allowed_non_write_users: "*"
+          claude_args: |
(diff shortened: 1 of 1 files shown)


0xSolace:
Superseded by #233 which fixes all the same issues plus the root cause (multi-line YAML breaking --allowedTools).

PR #254, opened 2026-02-14
ci: harden agent-review workflow against shell injection

lawyered0:
## Summary
- harden agent-review workflow against shell command injection from untrusted PR metadata
- stop interpolating PR title/base/scope directly into shell run scripts
- move PR metadata to env vars and use quoted printf/file reads so untrusted content is treated as data

## Validation
- YAML parse check passed (ruby/Psych)

## Context
This is the approved follow-up from the critical bug review on workflow command injection risk.

### DIFF
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -263,6 +263,7 @@ jobs:
             printf -- "## PR Review: %s\n" "$PR_TITLE"
             echo ""
             echo "_Reviewed by Codex 5.3_"
+            echo "$AGENT_REVIEW_MARKER"
             echo ""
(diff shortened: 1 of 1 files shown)


lawyered0:
Maintainer unblock needed: this PR contains the Agent Review workflow fix required to stop the recurring `codex review --base ... [PROMPT]` parse-error comments on `pull_request_target` runs.

Current status: downstream PR #256 has all functional CI checks passing; only `review-pr` fails because base `develop` still runs the old workflow.

Please merge this PR into `develop` (or cherry-pick commit `44fdf64`) with maintainer/admin rights.

PR #268, opened 2026-02-15
milaidy: harden MCP stdio config validation

lawyered0:
## Summary
- add centralized MCP server config validation for stdio/http/sse server definitions
- restrict stdio commands to an allowlist and require bare executable names
- block inline execution flags (including attached short-option forms like `-c...`) for interpreters/package runners
- block dangerous env overrides and enforce strict types for args/env/cwd/timeoutInMillis
- apply validation to all MCP write surfaces: PUT /api/config, POST /api/mcp/config/server, PUT /api/mcp/config
- add dedicated regression tests for command/flag/env/server-name rejection behavior

## Validation
- bunx vitest run src/api/server.mcp-config-validation.test.ts src/api/server.websocket-auth.test.ts
- bun run check


0xSolace:

PR #270, opened 2026-02-15
database: block mutating sequence functions in read-only queries

lawyered0:
## Summary
- extend read-only SQL guard to reject mutating sequence functions (`nextval`, `setval`)
- keep existing keyword-based mutation detection and semicolon guard behavior
- add regressions for both bypass forms in read-only mode

## Testing
- bunx vitest run src/api/database.readonly-query-guard.test.ts
- bunx vitest run src/api/database.security.test.ts
- bun run check


0xSolace:

PR #271, opened 2026-02-15
runtime: block DNS-alias SSRF in custom action HTTP handler

lawyered0:
## Summary
- harden custom action HTTP SSRF guard with DNS resolution of hostnames
- block requests when any resolved address is loopback, private, link-local, metadata, or otherwise unsafe
- keep explicit allow for localhost API calls on the configured API port
- add regression coverage for nip.io alias bypasses and allow-cases

## Testing
- bunx vitest run src/runtime/custom-actions.test.ts
- bun run check

### DIFF
--- a/src/runtime/custom-actions.test.ts
+++ b/src/runtime/custom-actions.test.ts
@@ -100,4 +100,29 @@ describe("custom action SSRF guard", () => {
     expect(result.output).toBe("ok");
     expect(fetchSpy).toHaveBeenCalledTimes(1);
   });
+
+  it("blocks redirect responses and uses manual redirect mode", async () => {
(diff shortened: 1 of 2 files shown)


lawyered0:
One thing to note: the guard is DNS-resolution based (resolve, then fetch), so it’s still theoretically vulnerable to DNS rebinding / TOCTOU between lookup and the outbound request. That may be totally acceptable for our threat model (block obvious internal/metadata targets), but it’s worth noting. If we ever need stronger guarantees, we’d want to either pin the resolved IP for the actual request (no follow-redirect, connect-to-IP), or re-validate on redirects and potentially on connection.

PR #272, opened 2026-02-15
api: block SSRF targets in knowledge URL imports

lawyered0:
## Summary
- add DNS-aware SSRF validation for URL imports in `/api/knowledge/documents/url`
- enforce `http`/`https` protocol-only URL imports
- block loopback/private/link-local/metadata targets (including DNS alias resolutions)
- fail closed when hostname resolution fails
- add regression coverage for direct loopback block, DNS-alias block, and public host allow-case

## Testing
- bunx vitest run src/api/knowledge-routes.test.ts
- bun run check


0xSolace:

PR #273, opened 2026-02-15
api: restrict ollama import URL to loopback hosts

lawyered0:
## Summary
- add route-level validation for optional `ollamaUrl` in `POST /api/training/models/:id/import-ollama`
- require `ollamaUrl` to be a valid `http://` or `https://` URL
- restrict `ollamaUrl` hosts to loopback only (`localhost`, `127.0.0.1`, `::1`, including bracketed IPv6 form)
- prevent redirect-based host escape in Ollama import by using `fetch(..., { redirect: "manual" })`

## Tests
- `bunx vitest run src/api/training-routes.test.ts`
- `bunx vitest run src/services/training-service.import-ollama.test.ts`
- `bun run check`

## Regression Coverage
- rejects non-loopback `ollamaUrl` values (e.g. metadata IP)
- rejects unsupported protocols (e.g. `file://`)
- accepts bracketed IPv6 loopback (`http://[::1]:11434`)
- verifies Ollama import fetch uses manual redirect mode

### DIFF
--- a/src/api/training-routes.test.ts
+++ b/src/api/training-routes.test.ts
@@ -319,6 +319,24 @@ describe("training routes", () => {
     expect(trainingService.importModelToOllama).not.toHaveBeenCalled();
   });
 
+  test("accepts bracketed IPv6 loopback ollamaUrl", async () => {
+    const result = await invoke({
+      method: "POST",
+      pathname: "/api/training/models/model-1/import-ollama",
+      body: {
+        ollamaUrl: "http://[::1]:11434",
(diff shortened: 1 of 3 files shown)


lawyered0:
Follow-up hardening:

- Explicit IPv6 loopback coverage (including bracketed form) with regression test.
- Switched Ollama import fetch to redirect: 'manual' to prevent redirect-based SSRF escape.
- All gates green (bun run check + targeted suites).

PR #276, opened 2026-02-15
security: disable WebSocket query-token auth by default

0xVera:
## Summary
Turns WebSocket query-parameter token auth into an explicit opt-in.

## Changes
- Query token auth is now disabled by default
- Enable only when `MILAIDY_ALLOW_WS_QUERY_TOKEN=1`
- Add tests for default rejection and explicit opt-in acceptance

## Security impact
Reduces token leakage risk from URLs (logs, browser history, intermediaries) while preserving backward compatibility via explicit opt-in.

### DIFF
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -3198,6 +3198,9 @@ export function resolveWalletExportRejection(
 }
 
 function extractWsQueryToken(url: URL): string | null {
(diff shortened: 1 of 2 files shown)


Dexploarer:
Cherry-picked onto develop as d36eb659

PR #287, opened 2026-02-15
runtime: block full IPv6 link-local range in custom action SSRF guard

lawyered0:
## Summary
- harden custom action HTTP SSRF guard to block full IPv6 link-local `fe80::/10`
- replace `^fe80:` with `^fe[89ab][0-9a-f]:` in blocked IP patterns
- add regression coverage for direct `http://[fea0::1]/...` target to ensure request is blocked before fetch

## Validation
- bunx vitest run src/runtime/custom-actions.test.ts
- bun run check *(fails only on pre-existing baseline files: .github/trust-scoring.cjs and .github/contributor-trust.json)*

### DIFF
--- a/src/runtime/custom-actions.test.ts
+++ b/src/runtime/custom-actions.test.ts
@@ -67,6 +67,17 @@ describe("custom action SSRF guard", () => {
     expect(fetchSpy).not.toHaveBeenCalled();
   });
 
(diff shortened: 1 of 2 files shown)


lawyered0:
Superseded by #291. This replacement preserves the same minimal fix and regression coverage on current develop. (Force-push to the existing branch is blocked by repo policy in this environment.)

PR #428, opened 2026-02-21
app: improve chat send and image remove controls

lawyered0:
## Summary\n- disable chat Send when input is empty/whitespace so UI matches submit behavior\n- improve pending-image remove control usability on touch/mobile and keyboard focus\n- add ChatView tests covering both behaviors\n\n## Testing\n- bunx vitest run apps/app/test/app/chat-view.test.tsx


lalalune:
Changes already present on develop. Closing.

PR #455, opened 2026-02-22
app: fix save command modal keyboard UX regression

lawyered0:
## Summary\n- fix keyboard interaction regression in Save Command modal\n- close modal on Escape only (not Enter/Space) to avoid accidental dismissal while typing\n- add focused tests for Enter no-close behavior and Escape close behavior\n\n## Classification\n- UX regression fix (P2)\n\n## Validation\n- bunx vitest run apps/app/test/app/save-command-modal.test.tsx apps/app/test/app/chat-view.test.tsx


lalalune:

PR #456, opened 2026-02-22
app: harden palette and save modal keyboard UX

lawyered0:
## Summary\n- prevent invalid command palette navigation indices when no results exist\n- clamp command palette active index when results shrink\n- fix save-command modal key handling so Enter/Space no longer close the dialog globally\n- add IME composition guard and ARIA label/error wiring in save-command modal\n- add focused app tests for both components and track execution in tasks/todo.md\n\n## Testing\n- bun run --cwd apps/app test -- test/app/command-palette.test.tsx\n- bun run --cwd apps/app test -- test/app/command-palette.test.tsx test/app/save-command-modal.test.tsx


lalalune:

PR #457, opened 2026-02-22
app: prevent accidental secrets picker close on Enter/Space

lawyered0:
## Summary\n- fix critical secrets picker keyboard behavior: modal now closes only on Escape\n- prevent accidental modal close on Enter/Space during picker interaction\n- add focused regression test for Enter/Space/Escape behavior\n- document plan and verification in tasks/todo.md\n\n## Testing\n- bun run --cwd apps/app test -- test/app/secrets-view.test.tsx


lalalune:

PR #498, opened 2026-02-23
docs: add Ollama setup guide with OpenAI-compatible workaround

sailorpepe:
## What

Adds a "Using Ollama" section to the README under Model Providers.

## Why

The `@elizaos/plugin-ollama` has an SDK version incompatibility with AI SDK v6 (see elizaos-plugins/plugin-ollama#18). Users who select Ollama during onboarding hit an `Unsupported model version v1` error with no obvious fix.

This documents the working workaround: routing Ollama through its OpenAI-compatible endpoint (`localhost:11434/v1`), which uses the functional `@elizaos/plugin-openai` instead.

## Changes

- Added Ollama installation instructions
- Documented the known plugin-ollama SDK incompatibility with a warning callout
- Added the `milady.json` configuration for the OpenAI-compatible workaround
- Added a recommended models table for local use

## Testing

Verified this configuration works with `gemma3:4b` and `llama3.2` on macOS (Apple Silicon). Agent loads 14/14 plugins and responds correctly via the dashboard chat.

---

*tested by a human. that's the split.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,44 @@ Or use `~/.milady/.env` for secrets.
 | [xAI](https://x.ai) | `XAI_API_KEY` | grok, based |
 | [DeepSeek](https://deepseek.com) | `DEEPSEEK_API_KEY` | reasoning arc |
 
+### Using Ollama (local models)
+
+[Ollama](https://ollama.ai) lets you run models locally with zero API keys. Install it, pull a model, and configure Milady:
+
+```bash
+# install ollama
+curl -fsSL https://ollama.ai/install.sh | sh
+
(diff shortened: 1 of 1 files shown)


sailorpepe:
Hey! First-time contributor here 👋 The auto-reviewer skipped this since I'm external (`sailorpepe does not have write permissions`). This is a docs-only change — adds Ollama setup instructions with the OpenAI-compatible workaround for the plugin-ollama SDK incompatibility (related to elizaos-plugins/plugin-ollama#18). Would appreciate a human review when you get a chance 🙏

PR #503, opened 2026-02-23
docs(ollama): add troubleshooting section with SDK workaround and failure guidance

sailorpepe:
## What

Adds a **Troubleshooting** section to `docs/plugin-registry/llm/ollama.md` covering setup failures and recovery steps.

## Why

Partially addresses **MW-11** (#476) — the Ollama integration is missing setup, failure, and verify guidance. Users who select Ollama during onboarding hit a cryptic `Unsupported model version v1` error with no documented fix. This is the most common failure path for local model users.

## Changes

Added three troubleshooting entries to the existing Ollama plugin doc:

1. **\"Unsupported model version v1\" error** — root cause (SDK v1/v6 mismatch) and the OpenAI-compatible endpoint workaround with `milady.json` config example
2. **Ollama Not Detected** — checklist for verifying the Ollama server, models, and port config
3. **Slow Responses** — RAM/VRAM debugging and model sizing guidance

Also links to the upstream tracking issue: elizaos-plugins/plugin-ollama#18

## Testing

Verified the workaround config works with `gemma3:4b` and `llama3.2` on macOS Apple Silicon. Agent loads 14/14 plugins and responds correctly.

---

*tested by a human. that's the split.*

### DIFF
--- a/docs/plugin-registry/llm/ollama.md
+++ b/docs/plugin-registry/llm/ollama.md
@@ -146,6 +146,52 @@ export OLLAMA_BASE_URL=http://192.168.1.100:11434
 
 Secure with a reverse proxy (Nginx + TLS) for production.
 
+## Troubleshooting
+
+### "Unsupported model version v1" Error
+
+**Symptoms:** Agent crashes or silently fails on first chat. Terminal shows:
+
+```
+Error: Unsupported model version v1
+```
+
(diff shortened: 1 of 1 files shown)


lalalune:
Changes have been squashed and merged into develop via PR #518.

PR #769, opened 2026-03-03
release: merge develop into main

lalalune:
Release sync from `develop` into `main` with the current production app filter hardening included.

Follow-up fixes in this update:
- use an exact production app allowlist instead of substring matching
- keep non-production app visibility unrestricted
- add focused coverage for the filter behavior


Dexploarer:
@claude review this pr against the workspaces actions, reviews, and trust scoring. address any gatekeepers,issues, and comments that arise.

PR #786, opened 2026-03-03
fix(deps): add missing @elizaos/plugin-pi-ai dependency

dutchiono:
## Description
Fixes a silent 500/crash during \startApiServer\ caused by \@elizaos/plugin-pi-ai\ not resolving.

Without this dependency, \server.js\ fails to bind to port 2138 because the module-level import throws an \ERR_MODULE_NOT_FOUND\ before the server is initialized.

## Testing
- Tested in local environment.
- Verified \milady start\ boots the API server correctly after applying this patch.


odilitime:
needs to be next though, do we need to publish it on npm?

PR #856, opened 2026-03-08
fix(windows): fix 4 electrobun startup bugs blocking Windows canary

dutchiono:
## Summary

Investigated by running the `2.0.0-alpha.80/81` canary build locally on Windows x64. All bugs reproduced and root-caused hands-on.

- **Bug 1** — `version.json` missing → silent crash before `main()` runs. The CI never writes this file for Windows; `getVersionInfo()` throws (not just logs) when it's absent. Fixed by adding post-build steps to write it from known build metadata.
- **Bug 2** — `renderer/` directory missing from package → blank window / no UI. Vite builds to `apps/app/dist/` but nothing copies it to `apps/app/electrobun/renderer/` before `electrobun build` runs. Fixed by adding a staging step in CI.
- **Bug 3** — `setupApplicationMenu()` called before `createMainWindow()` → permanent deadlock on Windows. CEF's Win32 message loop isn't running until the first window exists; the synchronous FFI call blocks forever waiting for the UI thread. Fixed by reordering in `src/index.ts` and guarding macOS-only menu roles behind `process.platform !== 'win32'`.
- **Bug 4** — `smoke-test-windows.ps1` `ParserError` → CI fails before app even runs. `$ErrorActionPreference = "Stop"` was placed before `param()`, which PowerShell requires to be the first statement. Also fixes the startup log path (`$env:APPDATA\Milady` not `$env:USERPROFILE\.config\Milady`). Rewrote the script based on the more complete version from `test/electrobun-cross-platform` with multi-port health checking, installer-zip fallback, and early failure detection.

**Bug 6 (documented, not fixed here):** `new BrowserWindow()` also deadlocks on Windows because `libNativeWrapper.dll` falls back to a shared-memory polling bridge instead of direct JS callbacks. Since the synchronous FFI call blocks the JS event loop, the native side can never get a response. This requires an Electrobun upstream fix — see `WINDOWS_ELECTROBUN_FIXES.md` for full root cause analysis and reference to blackboardsh/electrobun#250.

## Files changed

| File | Change |
|------|--------|
| `.github/workflows/release-electrobun.yml` | Stage renderer + inject `version.json` post-build |
| `apps/app/electrobun/src/index.ts` | Move `setupApplicationMenu()` after `createMainWindow()`; guard Windows |
| `apps/app/electrobun/scripts/smoke-test-windows.ps1` | Fix `param()` order; fix log path; expand diagnostics |
| `WINDOWS_ELECTROBUN_FIXES.md` | Full bug report with root causes and verification steps |

## Test plan

- [ ] CI Windows build no longer exits with code 1 before `main()` runs
- [ ] `Resources/version.json` present in packaged artifact
- [ ] `Resources/app/renderer/index.html` present in packaged artifact
- [ ] App reaches `[Main] Starting Milady (Electrobun)...` without hanging
- [ ] Smoke test PowerShell script parses without error
- [ ] macOS build unaffected

### DIFF
--- a/WINDOWS_ELECTROBUN_FIXES.md
+++ b/WINDOWS_ELECTROBUN_FIXES.md
@@ -120,6 +120,48 @@ So the diagnostic output on failure would be empty even when logs exist.
 
 ---
 
+## Bug 6 — `new BrowserWindow()` deadlocks on Windows → native window never appears
+
+**Symptom:** After fixes 1–3, the app logs progress further:
+```
+[Main] Starting Milady (Electrobun)...
+[WebGPU] Native Dawn runtime ready at ...
+[Renderer] Static server on http://127.0.0.1:5174
+```
+Then hangs for ~10 seconds inside `new BrowserWindow({...})` and exits.
+No native window ever appears. No CEF helper processes (`bun Helper.exe`,
+`bun Helper (GPU).exe` etc.) are spawned.
+
+**Root cause:** `app.log` always shows:
+```
+setJSUtils called but using map-based approach instead of callbacks
+```
+This means the Electrobun native library (`libNativeWrapper.dll`) could not
+establish direct callback pointers into Bun's JS runtime on Windows, so it
+fell back to a shared-memory polling mechanism.
+
+`new BrowserWindow()` makes a **synchronous** FFI call. The native code
+creates the CEF window and then needs to signal JS that it's ready. With the
+**callbacks** approach this works — native calls a JS function pointer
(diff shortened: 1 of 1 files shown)


Dexploarer:
Closing as superseded by electrobun@392e6095. The clean fixes from this PR that were still missing on current Electrobun were landed directly on the branch: renderer staging, packaged version.json injection, and the Windows %APPDATA% startup-log path in smoke-test-windows.ps1. The remaining stale Electron-side scope in this fork PR is not being merged into the Electrobun branch.

PR #864, opened 2026-03-08
milady: fix companion ui and plugin setup links

Dexploarer:
## Summary
- fix companion UI regressions across settings, stream navigation, VRM loading, and shell close flows
- classify Retake as a streaming plugin and add setup/help links across plugin surfaces
- add regression coverage for plugin metadata, wallet actions, stream popout behavior, VRM remount loading, and settings flows

## Testing
- bun run pre-review:local

### DIFF
--- a/apps/app/test/app/inventory-bsc-view.test.ts
+++ b/apps/app/test/app/inventory-bsc-view.test.ts
@@ -362,13 +362,26 @@ describe("InventoryView BSC-first", () => {
         node.type === "button" &&
         node.props["data-testid"] === "wallet-token-preflight",
     )[0];
+    const tokenInput = tree?.root.findAll(
+      (node) =>
+        node.type === "input" &&
+        node.props["data-testid"] === "wallet-quick-token-input",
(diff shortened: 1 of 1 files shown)


Dexploarer:
Pushed `b98c3677` on top of the earlier BSC quote-path fix.

This addresses the remaining failing unit test on the fresh head:
- align `inventory-bsc-view` with the shared BSC quote validation path by setting a valid token before asserting toolbar quote behavior
- preserves the production behavior change from `7d19fce1`, where toolbar quote now goes through the shared preflight-aware request path

Targeted validation used locally:
- `bunx vitest run --config vitest.unit.config.ts apps/app/test/app/inventory-bsc-view.test.ts apps/app/test/app/bsc-trade-panel.test.tsx`
- `bunx @biomejs/biome check apps/app/src/components/BscTradePanel.tsx apps/app/test/app/inventory-bsc-view.test.ts apps/app/test/app/bsc-trade-panel.test.tsx`

PR #1180, opened 2026-03-20
fix(onboarding): localhost cloud connect + scoped reset signals

dutchiono:
## Summary
- fix localhost onboarding cloud login by making app-core treat a same-origin available API backend as present during cloud auth routing
- keep the fix in dependency patch flow (scripts/patch-deps.mjs) instead of duplicating runtime globals in app HTML/entry
- scope /api/agent/reset to clear onboarding-completion signals only (dminEntityId, workspace) while preserving other gents.defaults settings

## Why
- resolves local onboarding Cloud Connect failures on localhost where app-core previously bypassed backend cloud routes
- addresses review concerns about duplicated/unvalidated API-base injection and over-broad defaults clearing

## Validation
- unx vitest run src/api/server.onboarding-status.test.ts src/api/cloud-routes.test.ts
- 
ode scripts/patch-deps.mjs

### DIFF
--- a/apps/app/electrobun/src/deferred-agent-startup.test.ts
+++ b/apps/app/electrobun/src/deferred-agent-startup.test.ts
@@ -40,7 +40,9 @@ describe("deferred agent startup (desktop)", () => {
   });
 
   it("initializes the embedded API token before renderer injection", () => {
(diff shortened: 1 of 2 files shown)


dutchiono:
Closing per request to stop this thread and avoid further churn.

PR #1346, opened 2026-03-25
[fix] fix desktop companion header glass bleed

SYMBaiEX:
## Summary
This fixes a desktop header regression where the companion glass shell styling leaked into the standard desktop header. On desktop, the top navigation was still wrapped in the same blurred glass bar intended for companion and character-facing header chrome, which made the full-width desktop header look like the companion overlay instead of the normal desktop surface.

## Root Cause
The shared `Header` component rendered the same outer shell treatment in both branches of its header chrome logic. Even when the active shell view was `desktop`, the wrapper still kept the blurred background, border, ring, and shadow classes, so the desktop header never actually dropped back to a transparent container.

## Fix
The desktop header path now uses a transparent shell wrapper, while the companion and character paths continue to use the dedicated glass treatment. This keeps the companion bar scoped to companion-oriented views and preserves the expected desktop header appearance without changing the separate `CompanionHeader` component.

## Validation
I ran the focused header test suite:

- `bun test packages/app-core/src/components/Header.test.tsx`

I also updated the header regression test so desktop explicitly asserts that the wrapper remains transparent and unblurred.

### DIFF
--- a/packages/app-core/src/components/Header.test.tsx
+++ b/packages/app-core/src/components/Header.test.tsx
@@ -478,7 +478,7 @@ describe("Header", () => {
     ).toHaveLength(1);
   });
 
-  it("renders the glassmorphic pill container in all modes", async () => {
+  it("keeps desktop header chrome transparent so companion glass stays scoped", async () => {
     const mockUseApp = {
       elizaCloudEnabled: false,
       elizaCloudConnected: false,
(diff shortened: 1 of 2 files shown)


Dexploarer:
Superseded by #1348 — lint/format fixed, applied cleanly. Thanks @SYMBaiEX!

PR #1367, opened 2026-03-26
fix: trajectory step ID mismatch causing synthetic LLM call data

HaruHunab1320:
startStep() generated disconnected step-xxx IDs, so LLM calls logged to one trajectory while the UI queried another — making all calls appear as synthetic backfills. Fix: return the trajectory ID from startStep(). Also delegate trajectory creation to the plugin's MESSAGE_RECEIVED handler instead of duplicating it in server.ts.


HaruHunab1320:
Folded into #1369 which includes both the ElizaCloud coding credentials fix and the trajectory step ID fix + core bump to alpha.108.

PR #1388, opened 2026-03-27
fix(release): publish plain canary windows installer artifact

dutchiono:
## Summary
- add a public canary Windows installer artifact that contains the plain Milady-Setup-*.exe
- keep the existing .exe.zip transfer artifact path for internal workflow reliability
- add workflow drift coverage and release-check assertions for the new canary artifact path

## Testing
- bun run release:check
- bunx vitest run scripts/electrobun-release-workflow-drift.test.ts
- bun run lint


Dexploarer:
Superseded by #1398, which rebases the same release workflow change onto current develop without the stale docs-frontmatter CI noise.

PR #1389, opened 2026-03-27
fix(ui): fix shell toggle symmetry and inactive state

rndrntwrk:
## Summary
- Remove per-button edge rounding (`rounded-l-xl`/`rounded-r-xl`/`rounded-none`) that caused the active highlight to clip flush against the container border
- All segments now use uniform `rounded-[10px]` inside a `rounded-[14px]` container with `p-1` padding — equal inset on all sides regardless of active position
- Give inactive segments a subtle `bg-white/[0.03]` baseline so the toggle reads as a unified control

## Test plan
- [ ] Click through all 3 shell views (companion, character, desktop)
- [ ] Verify active pill has equal inset from container border in all positions (left, middle, right)
- [ ] Verify inactive pills have subtle background presence
- [ ] Verify hover states still work on inactive pills
- [ ] Verify no layout shift when switching views
- [ ] Test on mobile viewport — verify responsive behavior unchanged

### DIFF
--- a/packages/app-core/src/components/companion/ShellHeaderControls.tsx
+++ b/packages/app-core/src/components/companion/ShellHeaderControls.tsx
@@ -244,21 +244,15 @@ export function ShellHeaderControls({
           aria-label="Switch shell view"
         >
           <legend className="sr-only">Switch shell view</legend>
-          {shellOptions.map(({ view, label, Icon }, index) => {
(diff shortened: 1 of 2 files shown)


Dexploarer:
Superseded by #1397, which rebases the shell toggle styling change onto current develop and adds the missing focused regression tests required by pre-review.

PR #1424, opened 2026-03-27
fix(railway): patch runtime deps for GPU probe + participant insert

dutchiono:
## Summary\n- patch dependency rewrite logic in scripts/patch-deps.mjs for Railway runtime stability\n- patch @elizaos/plugin-local-embedding to treat missing lspci in slim Linux images as a non-fatal skipped probe\n- patch @elizaos/plugin-sql participant insertion paths to avoid fragile ON CONFLICT behavior by using existence-check then insert\n\n## Why\nRailway deploy logs showed:\n- noisy Linux GPU detection failures due to /bin/sh: lspci: not found\n- participant insert failures: 	here is no unique or exclusion constraint matching ON CONFLICT specification\n\nThis change keeps startup clean and avoids failed participant writes in affected Postgres deployments.\n\n## Validation\n- ran 
ode scripts/patch-deps.mjs and confirmed patched targets are rewritten in installed plugin bundles

### DIFF
--- a/scripts/patch-deps.mjs
+++ b/scripts/patch-deps.mjs
@@ -361,7 +361,9 @@ patchTrajectoryLoggerJsonArrayDecode();
  */
 function patchLocalEmbeddingLinuxGpuProbe() {
   const relPaths = ["dist/index.js"];
-  const searchDirs = [resolve(root, "node_modules/@elizaos/plugin-local-embedding")];
+  const searchDirs = [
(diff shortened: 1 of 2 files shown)


dutchiono:
All CI/test/build checks are green on the latest commit (b7d6cf2). The only failing gate is **Agent Review Verdict**, and the bot comment indicates a temporary reviewer service outage (Claude + Codex fallback unavailable).\n\nCould a maintainer re-run the Agent Review workflow/check so this can merge?

PR #1441, opened 2026-03-28
fix(steward): resolve TypeScript errors in steward wallet UI

0xSolace:
Fixes all TypeScript compilation errors in the steward wallet UI code:

- Import `StewardPendingApproval`, `StewardTxRecord`, `StewardTxStatus` from `@miladyai/shared/contracts/wallet` (missing from agent contracts)
- Export `StewardView` from `app-shell-components` barrel file
- Fix `setActionNotice` signature in `ApprovalQueue` and `TransactionHistory` to match `AppState` type (`tone?` union instead of plain `string`)
- Fix Bun subprocess type incompatibility with type assertion in `steward-sidecar.ts`
- Fix possibly-null `healthCheckAbort` by capturing local reference

`npx tsc --noEmit` passes with zero errors.


0xSolace:
fixes merged into #1440 instead

PR #1444, opened 2026-03-28
fix(wallet): consolidate duplicate token sort controls

jqmwa:
## Summary
- Consolidate token sorting into a single control surface in the wallet rail
- Remove the unreachable refresh branch that was left behind after the sort controls were unified
- Keep wallet regression coverage aligned with the single-chain wallet flow

## Why
- Recent wallet changes exposed the same token sort state in multiple places, which made the active sorting entrypoint ambiguous
- This narrows the wallet header to the state that is still functional and exercised by the existing inventory flows

## Test plan
- [x] bun vitest packages/app-core/src/components/__tests__/InventoryView.wallet-settings.test.tsx packages/app-core/test/app/inventory-bsc-view.test.ts
- [x] bun run lint


Dexploarer:
@codex review

PR #1447, opened 2026-03-28
fix(ui): align onboarding language selector with app header

rndrntwrk:
## Summary
- remove the onboarding-specific language trigger override that shrank and re-skinned the selector
- reuse one shared language trigger class for both onboarding and the in-app header
- tighten onboarding test coverage so the dropdown props are asserted instead of hidden behind a dumb stub

## Testing
- /Users/mac/.bun/bin/bunx vitest run packages/app-core/src/components/OnboardingWizard.test.tsx packages/app-core/src/components/Header.test.tsx packages/app-core/src/components/companion/ShellHeaderControls.test.tsx
- git diff --check

## Notes
- scope is limited to selector parity only
- no language-switch behavior, menu behavior, or i18n logic changed

### DIFF
--- a/packages/app-core/src/components/Header.test.tsx
+++ b/packages/app-core/src/components/Header.test.tsx
@@ -28,6 +28,8 @@ vi.mock("@miladyai/app-core/navigation", () => ({
 }));
 
 vi.mock("@miladyai/app-core/components", () => ({
+  LANGUAGE_DROPDOWN_TRIGGER_CLASSNAME:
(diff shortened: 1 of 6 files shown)


Dexploarer:
I couldn’t push follow-up fixes back to `rndrntwrk/milaidy` from this environment because the contributor fork rejected writes. I published the PR head plus the CI fixes to `milady-ai/milady` as [#1449](https://github.com/milady-ai/milady/pull/1449) so the change can rerun on a writable branch.

PR #1481, opened 2026-03-28
fix: skip Privy message when Steward wallet connected

0xSolace:
When Steward is connected and has wallet addresses, skip the 'No onchain wallet, log in with Privy' message. Also updates copy to be steward-aware.


Dexploarer:
@claude

PR #1492, opened 2026-03-29
fix: log warnings in catch blocks that silently return empty arrays

awidearray:
## LARP Assessment — Finding (4): Error handling that silently swallows failures

> *"Check for error handling that silently swallows failures."*

Four services had `catch { return []; }` or `catch { return {}; }` blocks that made real errors **indistinguishable from "no data"**:

| File | Function | Error becomes |
|------|----------|--------------|
| `registry-client.ts` | `getConfiguredEndpoints()` | Config load fails → `[]` (looks like "no custom endpoints") |
| `skill-marketplace.ts` | `readInstallRecords()` | Corrupt JSON → `{}` (looks like "no installs") |
| `custom-actions.ts` | `loadCustomActions()` | Config load fails → `[]` (looks like "no custom actions") |
| `api/client.ts` | `listCodingAgentScratchWorkspaces()` | Network error → `[]` (looks like "no workspaces") |

An operator debugging "why aren't my custom actions loading?" has **zero signal** that the config file is malformed.

## Changes

- All four now `logger.warn()` / `console.warn()` before returning the fallback
- `skill-marketplace` additionally skips the warning for `ENOENT` (file doesn't exist yet is normal on first run)
- Fallback behavior preserved — no throwing, no UI crash

## Test plan

- [ ] `bun run test` passes
- [ ] Introduce a malformed `milady.json` config → verify warning appears in logs for registry-client, custom-actions
- [ ] Delete `installs.json` → verify skill-marketplace does NOT warn (ENOENT is expected)
- [ ] Corrupt `installs.json` → verify skill-marketplace DOES warn

Shaw hotkey: *"`catch {} return []` is gaslighting your future self."*

### DIFF
--- a/packages/agent/src/runtime/custom-actions.ts
+++ b/packages/agent/src/runtime/custom-actions.ts
@@ -16,7 +16,7 @@ import {
 import { request as requestHttps } from "node:https";
 import net from "node:net";
 import { Readable } from "node:stream";
-import type { Action, HandlerOptions, IAgentRuntime } from "@elizaos/core";
+import { type Action, type HandlerOptions, type IAgentRuntime, logger } from "@elizaos/core";
 import { loadElizaConfig } from "../config/config";
 import {
   resolveApiToken,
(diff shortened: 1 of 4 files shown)


Dexploarer:
Reviewed. The warning-on-fallback behavior is useful. I consolidated it into #1501 with focused coverage for the config, install-record, and client fallback paths. Closing as superseded.

PR #1495, opened 2026-03-29
fix: remove dead deprecated stubs with zero importers

awidearray:
## LARP Assessment — Finding (1): Stubbed functions that return fake data

> *"Check for stubbed functions that return fake data."*

Three exported functions existed solely as stubs returning hardcoded values, with **zero importers** anywhere in the codebase:

| Function | File | Returns | Importers |
|----------|------|---------|-----------|
| `isOfficialVrmIndex()` | `state/vrm.ts` | `false` always | 0 (only definition site) |
| `getVrmNeedsFlip()` | `state/vrm.ts` | `false` always | 0 (only definition site) |
| `ethToNumber` | `policy-controls/helpers.ts` | alias for `parseAmount` | 0 (only definition site) |

All three were marked `@deprecated` and re-exported via barrel files, but no consumer ever imported them. They inflated the public API surface and misled developers into thinking they served a purpose.

**Verification:** `rg "isOfficialVrmIndex|getVrmNeedsFlip"` and `rg "ethToNumber"` each return only the definition file.

## Changes

All three removed.

## Test plan

- [ ] `bun run build` succeeds (no broken imports)
- [ ] `bun run test` passes
- [ ] `bun run check` passes

Shaw hotkey: *"Dead code isn't legacy — it's noise pretending to be signal."*

### DIFF
--- a/packages/app-core/src/components/policy-controls/helpers.ts
+++ b/packages/app-core/src/components/policy-controls/helpers.ts
@@ -13,11 +13,6 @@ export function parseAmount(value: string): number {
   return Number.isNaN(n) ? 0 : n;
 }
 
-/**
- * @deprecated Use parseAmount instead.
- */
-export const ethToNumber = parseAmount;
-
 export function formatHour(h: number): string {
   if (h === 0) return "12:00 AM";
   if (h === 12) return "12:00 PM";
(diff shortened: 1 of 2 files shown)


Dexploarer:
Reviewed and not taking this. These are exported deep-path APIs, and removing them is public-surface churn for almost no gain. The repo does not need a breaking API cleanup PR here.

PR #1538, opened 2026-03-31
milady: add release-tagged CDN asset delivery

dutchiono:
## Summary
- add release-tagged static asset delivery for app and homepage
- validate managed CDN assets from release context with fork-safe checks
- offload blocked .spz world assets via tagged raw GitHub fallback
- harden desktop preflight for hoisted electrobun installs in CI
## Validation
- 	est-electrobun-release.yml passed on codex/jsdelivr-static-assets
- full release workflow is running on dutchiono/milady tag 2.0.0-alpha.131-cdn.4
- targeted vitest suites for CDN/release contracts and desktop preflight passed locally
- un run lint
## Notes
- fork release validation uses tagged raw GitHub for verification; canonical upstream production delivery remains jsDelivr-backed
- this PR is draft until the current fork release run finishes cleanly


0xSolace:
## Full Review — CDN Asset Delivery PR

### CDN Provider Evaluation

**Current**: jsDelivr (`cdn.jsdelivr.net/gh/milady-ai/milady@{tag}/...`)
**Recommendation**: Consider switching to **raw GitHub** as the primary CDN, or at minimum make the provider easily swappable.

| Criteria | jsDelivr | GitHub Raw | Cloudflare R2 |
|---|---|---|---|
| Setup | Zero | Zero | Moderate (needs R2 bucket + worker) |
| Speed | Fast global CDN | Slower, GitHub edge only | Fastest (you control it) |
| Reliability | Has had outages (2021, 2023) | If GitHub is up | You control uptime |
| File restrictions | Blocks `.spz` (already worked around) | None | None |
| Size limits | 50MB/file | 100MB/file | 5TB/file |
| Cost | Free | Free | ~$0.015/GB egress |
| External dependency | Yes (Cloudflare-backed but separate) | Only GitHub | Only Cloudflare (already set up) |

**The `.spz` fallback already proves raw GitHub works.** For static assets served at release time (not high-traffic CDN — these are desktop apps and self-hosted instances, not a public website with millions of hits), raw GitHub is simpler and sufficient:
- No file type restrictions
- No external service dependency beyond GitHub itself
- One URL pattern instead of two (jsDelivr + raw GitHub fallback)
- The code already has the `buildRawGitHubAssetBase()` function

If traffic ever demands a real CDN, Cloudflare R2/Pages is the right long-term play since you already have zones set up. But for now, raw GitHub simplifies the architecture and removes the jsDelivr dependency.

---

### Code Review — Detailed Findings

#### ✅ What's Good

1. **`resolveAppAssetUrl()` is well-designed** — handles absolute URLs, relative paths, leading-slash stripping, `file://` schemes, custom schemes (electrobun), and CDN base URL resolution. The `.spz` → raw GitHub fallback regex is correct.

2. **`resolveHomepageAssetUrl()` is clean** — separate function with appropriate scope, reads from release-data at build time.

3. **Static asset manifest** — deterministic (sorted paths, `path.sep` → `/` normalization), cross-platform safe. Drift detection works correctly.

4. **Pruning script** is safe — only runs post-build, only removes from `dist/` (not source), only when CDN is configured. Falls back gracefully to full local assets when no release tag is set.

5. **Retry logic in `validate-cdn-assets.mjs`** — handles jsDelivr cache propagation delay properly (16 attempts × 15s in CI). Only retries jsDelivr URLs, not raw GitHub.

6. **`boot-config.ts` path change** — dropping leading `/` from `compressedVrmPath`, `previewPath`, etc. is correct since `stripLeadingPathMarkers` in `resolveAppAssetUrl` would remove it anyway. More consistent.

7. **`patch-deps.mjs` `ROOT → root` fix** — correct bug fix. The variable is `const root = resolve(__dirname, "..")`, `ROOT` was undefined (would throw ReferenceError).

8. **Test coverage is solid** — `asset-cdn.test.ts`, `asset-url.test.ts`, `static-asset-manifest.test.ts`, and `homepage/asset-url.test.ts` cover the happy paths, edge cases (empty strings, already-absolute URLs, `.spz` routing, fork validation).

#### ⚠️ Issues to Address

1. **Docker build doesn't pass `MILADY_RELEASE_TAG`** — `.github/workflows/build-docker.yml` only passes `VERSION`, not `MILADY_RELEASE_TAG`. This means Docker builds won't trigger CDN mode. The image stays large (old behavior) but functional. Need to wire `MILADY_RELEASE_TAG` into the Docker build workflow for the size reduction to actually happen.

2. **`write-homepage-release-data.mjs` Windows compat** — removed `process.platform === "win32" ? "bunx.cmd" : "bunx"` and now uses bare `bunx` with `execFileSync`. On Windows, `execFileSync` without `shell: true` won't resolve `.cmd` extensions. Low risk (likely Linux-only CI), but inconsistent with `run-node.mjs` which still handles Windows.

3. **`parseBunVersion` refactor** — changed from "scan first line for semver anywhere" to "match first whitespace token as exact semver". If `bun --version` ever outputs `"Bun 1.3.10"` (prefixed), the new code breaks. In practice `bun --version` outputs just `"1.3.10"`, so this works today, but the old code was more defensive.

4. **`VrmEngine.playWaveGreeting()`** at line 2591 has a hardcoded `"animations/emotes/greeting.fbx"` that does NOT go through `resolveAppAssetUrl`. It's currently dead code (defined but never called externally), but if someone calls it in the future it would break in CDN mode. Should either route through the resolver or be removed.

5. **`resolveHomepageAssetUrl` uses `VITE_ASSET_BASE_URL`** (same env var as app) as override. If both app and homepage are built in the same process, they'd share the same override, but app and homepage have different CDN roots. This is fine because the override is a last-resort escape hatch, but the naming is confusing.

#### 📋 Nits

- `biome runner change` — switching from `stdio: "inherit"` to manual stdout/stderr capture + "No files were processed" skip is correct. This handles the case where biome exits non-zero for an empty file list. Good fix.

- The PR deletes the entire `benchmarks/` directory and `docs/core-audit-2026-03-30.md` — unrelated cleanup should be in a separate PR, but not blocking.

- `patch-deps.mjs` template literal encoding change (unicode escapes → backtick templates) is cosmetic but correct.

#### 🔒 Race Conditions

The CDN validation has appropriate retry logic for jsDelivr cache propagation. The `validateGroup` function batches HEAD requests and retries only failing URLs. The 16×15s (4 min) retry window in CI is reasonable for jsDelivr, which typically propagates within 1-2 minutes after a tag is pushed.

No race conditions in the asset resolution at runtime — `resolveAppAssetUrl` is synchronous and reads from `getBootConfig()` which is set once at startup.

---

### Summary

The architecture is sound. The CDN approach works correctly with proper fallbacks. Main actionable items:
1. Wire `MILADY_RELEASE_TAG` into Docker build workflows (or the size reduction never activates)
2. Consider raw GitHub over jsDelivr to eliminate the external dependency and `.spz` workaround
3. Fix or remove `VrmEngine.playWaveGreeting()` dead code

No blocking issues found. The PR is safe to merge as-is — the missing Docker workflow integration means it's effectively a no-op for Docker builds until that's added, which is the safe default.

0xSolace:
Hey @dutchiono — good foundation here. We've built on top of your CDN infrastructure with a few additions:

1. **Switched default CDN from jsDelivr → GitHub raw** (`raw.githubusercontent.com`). Removes the external dependency, eliminates the .spz workaround (raw GitHub serves all file types), and works automatically for forks via `GITHUB_REPOSITORY` env var. `buildJsDelivrAssetBase()` is kept as opt-in if anyone needs it.

2. **VRM lazy loading** — killed the 73MB eager preload on character-select, replaced with thumbnail-only preload (~640KB). Added an LRU ArrayBuffer cache in VrmEngine so avatar swaps are instant on cache hit.

3. **Cache headers** for VRM files served from the container.

Opened **#1541** which includes all your commits plus these additions, targeting develop. This PR can be closed in favor of that one.

Thanks for the solid scaffolding — the asset URL helpers, manifest system, and fork detection were all clean.

dutchiono:
Superseded by #1541 and the branch-sync follow-up in #1542.

PR #1543, opened 2026-03-31
chore: clean up CDN review notes

dutchiono:
Small follow-up to #1541.\n\nChanges:\n- add a focused regression test asserting validation URLs stay aligned with managed asset URLs\n\nValidation run locally:\n- un run lint\n- un run typecheck\n- unx vitest run scripts/asset-cdn.test.ts scripts/lib/desktop-preflight.test.ts\n\nNote: the branch was created from current develop after #1541 merged, so this stays isolated from the main CDN delivery changes.


dutchiono:
Optional follow-up test only; closing to keep the CDN rollout cleanup minimal.

PR #1636, opened 2026-04-04
fix: use env-configured port instead of hardcoded 31337 fallbacks

TSchonleber:
## What

Three related fixes where the API port was hardcoded to `31337` instead of reading from `MILADY_API_PORT` / `ELIZA_PORT` env vars.

## Why

When running on a non-default port (via `MILADY_API_PORT` override or dev-server auto-negotiation), these hardcoded values cause:

1. **Loopback API calls go to the wrong port** — `resolveCompatLoopbackApiBase()` falls back to `127.0.0.1:31337` when the `Host` header is missing. Functions like `clearCompatRuntimeStateViaApi()` silently fail.

2. **CORS rejects requests after port change** — `getCorsAllowedPorts()` cached the port set once on first request. After `syncResolvedApiPort()` updates env vars at runtime, the CORS allowlist is stale and rejects requests from the new UI port.

3. **Steward webhook registers on wrong port** — `tryRegisterStewardWebhook()` defaults to port `31337` instead of reading env.

## Changes

- **`server.ts:resolveCompatLoopbackApiBase`** — read `MILADY_API_PORT ?? ELIZA_PORT ?? "31337"` as fallback
- **`server.ts:getCorsAllowedPorts`** — remove stale cache, call `buildCorsAllowedPorts()` directly (trivial Set construction, no perf concern)
- **`steward-bridge.ts:tryRegisterStewardWebhook`** — read env vars for default port

## Testing

- All 16 existing CORS port allowlist tests pass
- `bun run build` succeeds

### DIFF
--- a/packages/app-core/src/api/server.ts
+++ b/packages/app-core/src/api/server.ts
@@ -256,7 +256,9 @@ function resolveCompatPgliteDataDir(config: ElizaConfig): string {
 function resolveCompatLoopbackApiBase(
   req: Pick<http.IncomingMessage, "headers">,
 ): string {
-  const host = req.headers.host?.trim() || "127.0.0.1:31337";
+  const host =
+    req.headers.host?.trim() ||
+    `127.0.0.1:${process.env.MILADY_API_PORT?.trim() || process.env.ELIZA_PORT?.trim() || "31337"}`;
   return `http://${host}`;
 }
 
@@ -581,12 +583,17 @@ export function buildCorsAllowedPorts(): Set<string> {
   return ports;
 }
 
-/** Lazily cached port set — computed once on first request. */
+/** Lazily cached port set — computed once, invalidated on port changes. */
 let _cachedCorsAllowedPorts: Set<string> | undefined;
 function getCorsAllowedPorts(): Set<string> {
   if (!_cachedCorsAllowedPorts)
     _cachedCorsAllowedPorts = buildCorsAllowedPorts();
   return _cachedCorsAllowedPorts;
+}
+
+/** Invalidate the cached CORS port set so it is recomputed on next request. */
+export function invalidateCorsAllowedPorts(): void {
+  _cachedCorsAllowedPorts = undefined;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/app-core/src/api/server.ts
```diff
@@ -581,12 +583,13 @@ export function buildCorsAllowedPorts(): Set<string> {
...
-let _cachedCorsAllowedPorts: Set<string> | undefined;
+/**
+ * Return the current set of allowed CORS ports.
+ * Re-reads env vars on every call so port changes from
+ * syncResolvedApiPort() are picked up immediately.
+ */
 function getCorsAllowedPorts(): Set<string> {
-  if (!_cachedCorsAllowedPorts)
-    _cachedCorsAllowedPorts = buildCorsAllowedPorts();
-  return _cachedCorsAllowedPorts;
```
odilitime, line 575: lost this optimization

odilitime (changes requested):

TSchonleber:
Updated — restored the cache with an explicit `invalidateCorsAllowedPorts()` export. Both `syncResolvedApiPort` call sites (dev-server.ts and eliza.ts) now invalidate the cache after port changes. Dynamic import to avoid circular deps.

Dexploarer (changes requested):
This PR still trusts `req.headers.host` as the primary loopback URL source in `resolveCompatLoopbackApiBase` — only the fallback changes to use env vars. PR #1641 correctly removes Host header trust entirely (replacing it with hardcoded `127.0.0.1` + resolved listener port).

#1641 should merge first as the security fix. After that, please rebase this PR on top of it — the `resolveCompatLoopbackApiBase` change will need to be dropped or adapted. The other changes (steward-bridge default port, CORS invalidation, dev-server port sync) are useful and can stay.

TSchonleber:
You're right — #1641 is the more complete fix that removes Host header trust entirely and uses the actual listener port. This PR's loopback changes are superseded by #1641. The remaining value here is the CORS cache invalidation and the steward webhook port fix.

Happy to rebase this on top of #1641 once it lands, or close it if you'd prefer to handle the CORS invalidation + steward port separately.

TSchonleber:
Since #1641 landed with the full SSRF + listener port fix, the loopback changes in this PR are now superseded. The remaining unique pieces are:

1. CORS cache invalidation (the cache is still not invalidated on port change)
2. Steward webhook port reading from env vars

Happy to open a focused follow-up PR for just the CORS invalidation if you want it, or close this one entirely. Your call.

PR #1644, opened 2026-04-04
fix(security): redact secrets from log output

TSchonleber:
## What

Two log hygiene fixes that prevent secrets from leaking via console output.

## Pairing code logged in full (MEDIUM)

`ensurePairingCode()` logged the full 12-character pairing code to `console.warn`:

```
[milady-api] Pairing code: A1B2-C3D4-E5F6 (valid for 10 minutes)
```

The pairing code can be exchanged for the **master API token** via `POST /api/auth/pair`. Anyone with log access (screen-watchers, log aggregators, crash reporters, terminal history) gets full API access.

**Fix:** Log only the first segment: `A1B2-****-****`

## API key length logged (LOW)

`credential-resolver.ts` logged key length at INFO level:

```
[credential-resolver] Resolved ANTHROPIC_API_KEY for anthropic (47 chars, api-key)
```

Key length narrows brute-force search space and confirms which providers have live credentials.

**Fix:** Remove the length from the log message.

## Testing

- `bun run build` succeeds

### DIFF
--- a/packages/app-core/src/api/auth-pairing-compat-routes.ts
+++ b/packages/app-core/src/api/auth-pairing-compat-routes.ts
@@ -74,8 +74,10 @@ function ensurePairingCode(): string | null {
   if (!pairingCode || now > pairingExpiresAt) {
     pairingCode = generatePairingCode();
     pairingExpiresAt = now + PAIRING_TTL_MS;
+    // Log only the first segment — enough to confirm the code was generated
+    // without exposing the full secret to log aggregators or screen-watchers.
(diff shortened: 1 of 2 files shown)


odilitime:
if you don't output how are you supposed to know it to pair?

TSchonleber:
Closing — the pairing code is intentionally displayed in the terminal so the user can enter it in the desktop app. Redacting it defeats the purpose. The console.warn output IS the user-facing display mechanism.

PR #1646, opened 2026-04-04
test: add coverage for server-config-filter (sensitive env stripping)

TSchonleber:
## What

Adds 7 tests for `filterConfigEnvForResponse` in `server-config-filter.ts`.

## Why

This function is a security boundary — it prevents private keys, API tokens, and database credentials from leaking via `GET /api/config` responses. It had zero test coverage.

## Tests

1. Strips all 11 sensitive env keys from response
2. Case-insensitive matching (evm_private_key, Evm_Private_Key both caught)
3. Does not mutate the original config object
4. Returns config as-is when env is missing
5. Returns config as-is when env is not an object
6. Returns config as-is when env is an array
7. Handles empty env object

All pass:
```
✓ packages/app-core/src/api/server-config-filter.test.ts (7 tests) 3ms
```

### DIFF
--- a/packages/app-core/src/api/server-config-filter.test.ts
+++ b/packages/app-core/src/api/server-config-filter.test.ts
@@ -1,0 +1,99 @@
+import { describe, expect, it } from "vitest";
+import {
+  filterConfigEnvForResponse,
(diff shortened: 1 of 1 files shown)


Dexploarer (approved):

PR #1649, opened 2026-04-04
test: add coverage for rate limiter utility

TSchonleber:
8 tests for the createRateLimiter used in auth pairing, wallet export guard, and other rate-limited endpoints. Covers: allow/block logic, window expiry, independent key tracking, peek vs check semantics, clear/dispose lifecycle, retryAfterSeconds accuracy.


Dexploarer (approved):

PR #1650, opened 2026-04-04
fix: gateway discovery removes wrong gateway due to overly broad OR matching

TSchonleber:
## What

`handleServiceLost()` in the Electrobun gateway discovery used `||` to match name, host, or port — any single field matching caused deletion.

## Bug

Two gateways sharing the same port (common: port 443, default gateway port) causes the wrong one to be removed when an unrelated service disappears. The loop `break`s immediately, so the correct gateway is never checked.

**Repro:** Discover two gateways on the same port but different hosts. When one goes down, the other may be removed from the map.

## Fix

Match on name AND host. Both must match for a gateway to be considered the same service that was lost.

### DIFF
--- a/apps/app/electrobun/src/native/gateway.ts
+++ b/apps/app/electrobun/src/native/gateway.ts
@@ -156,12 +156,16 @@ export class GatewayDiscovery extends EventEmitter {
   }
 
   private handleServiceLost(service: BonjourService): void {
+    // Build the same stableId that handleServiceFound uses so we match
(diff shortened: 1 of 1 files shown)


TSchonleber:
Updated — now matches on `stableId` (the map key) instead of individual fields. Builds the same `txt.id ?? name-host:port` composite that `handleServiceFound` uses. This avoids both the original OR-matching bug and the hostname-vs-IP mismatch the bot correctly identified.

Dexploarer (approved):

PR #1651, opened 2026-04-04
fix: add spawn error handler to repo setup to prevent unhandled exception

TSchonleber:
The child process spawned for each setup step only listens for `exit` but not `error`. If a script path doesn't exist (ENOENT), Node emits `error` which becomes an unhandled exception, crashing setup and leaving `.eliza-repo-setup.lock` orphaned — blocking all future installs until manually deleted.

Fix: add `child.on('error', reject)`. 3 lines.


Dexploarer (approved):

PR #1652, opened 2026-04-04
fix: prevent double FFmpeg spawn on stream restart (volume/mute change)

TSchonleber:
## Bug

`restart()` kills FFmpeg and starts a new instance but never sets `_intentionalStop=true` before killing. When FFmpeg's exit event fires asynchronously, the exit handler sees `_running=true` and `_intentionalStop=false`, so it calls `autoRestart()` **concurrently** with the manual restart. Two FFmpeg processes run simultaneously.

## Trigger

Call `setVolume()`, `mute()`, or `unmute()` while streaming.

## Fix

Set `_intentionalStop=true` before killing, reset to `false` before the manual `start()`. The exit handler now correctly skips `autoRestart()` during intentional restarts.

### DIFF
--- a/packages/agent/src/services/stream-manager.ts
+++ b/packages/agent/src/services/stream-manager.ts
@@ -160,6 +160,10 @@ class StreamManager {
     const savedStartedAt = this.startedAt;
     const savedFrameCount = this._frameCount;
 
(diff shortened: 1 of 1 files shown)


Dexploarer (approved):

PR #1653, opened 2026-04-04
fix: SSE stream parser drops multi-line data fields

TSchonleber:
## Bug

The SSE parser in `sendMessageStream()` overwrites `eventData` on each `data:` line:

```typescript
else if (line.startsWith("data: ")) eventData = line.slice(6); // overwrites!
```

Per the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), multiple `data:` lines in a single event should be concatenated with newlines. This causes silent data loss when Eliza Cloud sends multi-line SSE events.

## Fix

Append instead of overwrite: `eventData += (eventData ? "\n" : "") + line.slice(6)`

One-line change.


Dexploarer (approved):

PR #1654, opened 2026-04-04
fix: fire 'reconnecting' status before onDisconnect callback

TSchonleber:
## Bug

When cloud heartbeats fail past the threshold, ConnectionMonitor fires:
1. `onStatusChange('disconnected')`
2. `onDisconnect()`
3. `attemptReconnect()` → sets `onStatusChange('reconnecting')`

Callers see 'disconnected' and may show error UI or clear state, then immediately see 'reconnecting' — a confusing status flicker. The `onDisconnect` callback fires before any recovery attempt.

## Fix

Emit `reconnecting` first, then `onDisconnect()`. Callers see the transitional state and know recovery is in progress. The final `disconnected` only fires if all 10 reconnect attempts fail.

### DIFF
--- a/packages/agent/src/cloud/reconnect.ts
+++ b/packages/agent/src/cloud/reconnect.ts
@@ -70,7 +70,10 @@ export class ConnectionMonitor {
     );
 
     if (this.consecutiveFailures >= this.maxFailures) {
-      this.callbacks.onStatusChange?.("disconnected");
+      // Don't emit "disconnected" here — attemptReconnect() will emit
+      // "reconnecting" first, and only emits "disconnected" if all
+      // retry attempts fail. This avoids a misleading disconnected→
+      // reconnecting flicker for callers.
       this.callbacks.onDisconnect();
       await this.attemptReconnect();
     }


TSchonleber:
Updated — removed the duplicate emission from `tick()` entirely. Now `attemptReconnect()` is the sole owner of status transitions: emits `reconnecting` once at start, `connected` on success, `disconnected` only after all 10 retries fail. Single emission per state change.

Dexploarer (changes requested):
The PR removes the `onStatusChange("disconnected")` emission entirely and relies on `attemptReconnect()` to emit the correct status. However, if the retry loop exhausts all attempts, there is no visible `onStatusChange("disconnected")` call — callers would remain in a permanent `"reconnecting"` state.

Please verify that `attemptReconnect()` emits `"disconnected"` after exhausting all retries. If it does not, add `onStatusChange("disconnected")` as the final fallback after the retry loop.

TSchonleber:
The `disconnected` status IS emitted after all retries exhaust — it's at the end of `attemptReconnect()` (line 106 in the current develop branch):

```typescript
// After the for loop (10 attempts) falls through:
logger.error("[cloud-monitor] Failed to reconnect after 10 attempts");
this.reconnecting = false;
this.callbacks.onStatusChange?.("disconnected");  // ← here
```

The status flow is now:
1. Heartbeats fail past threshold → `onDisconnect()` fires (no status change yet)
2. `attemptReconnect()` starts → emits `reconnecting`
3a. Success → emits `connected`
3b. All 10 attempts fail → emits `disconnected`

Callers never get stuck in `reconnecting`. But this PR may already be superseded since develop has moved — happy to close if you've already handled this.

PR #1655, opened 2026-04-04
fix: remove double normalization of cloud base URL in CloudManager.init()

TSchonleber:
`normalizeCloudSiteUrl()` was called twice — once to normalize the URL, then again after validation before creating the client. The second call re-reads `ELIZAOS_CLOUD_BASE_URL` from `process.env`, so if that env var changes between calls (or differs from the config `baseUrl`), the validated URL differs from the one used for the client.

Fix: use the already-normalized `rawUrl` directly for client creation. Removes the dead `siteUrl` variable.


Dexploarer (approved):

PR #1685, opened 2026-04-05
fix: vincent oauth completes in external browser via server-side PKCE

Dexploarer:
## Summary

- The Vincent login flow never completed: users authenticated on heyvincent.ai, got redirected back, and were left on a blank tab while the desktop app spun on "connecting". Root cause: the PKCE `code_verifier` was stored in the desktop webview's `sessionStorage`, but the OAuth redirect lands in the user's external system browser (opened via `openExternalUrl` → `desktopOpenExternal` RPC), which has a **separate** `sessionStorage`. The callback effect saw a null verifier, silently bailed, and the desktop poll timed out.
- Move PKCE entirely server-side so the flow works regardless of which browser the redirect lands in.

## What changed

**New routes** (`packages/app-core/src/api/vincent-routes.ts`):

- `POST /api/vincent/start-login` — server registers with Vincent, generates `code_verifier` + `code_challenge` using `node:crypto`, stores a pending entry keyed by a fresh OAuth `state` UUID (10-min TTL, swept on access), and returns `{ authUrl, state, redirectUri }`. `redirect_uri` is built from `req.headers.host` so it matches the actual API origin in dev (`127.0.0.1:31337`) and prod (`127.0.0.1:2138`).
- `GET /callback/vincent` — looks up the pending entry by `state`, exchanges the code with Vincent, persists tokens to `ElizaConfig`, and returns a small self-closing HTML page. **Requires exact `state` match** — rejects requests with a missing or unknown state so no local process can forge a login for a session it didn't initiate. (PKCE already prevents cross-session token theft via Vincent's challenge/verifier check; state validation is the cheaper, clearer gate.)
- Token exchange now includes `redirect_uri` in the request body per **RFC 6749 §4.1.3** — Vincent was rejecting the exchange with `invalid_request: expected string, received undefined` without it, which is what was actually failing after the initial server-side move (found by tailing `[vincent/callback]` logs on dev).

**Routing** (`packages/app-core/src/api/server.ts`):

- `/callback/vincent` is wired through `handleVincentRoute` and **bypasses `ensureCompatApiAuthorized`** — the external browser has no compat token, and the PKCE verifier + state held server-side are the real credentials.

**Client** (`packages/app-core/src/api/client-vincent.ts` + `state/useVincentState.ts`):

- Adds `vincentStartLogin()` on `MiladyClient`.
- `useVincentState.handleVincentLogin` now just calls `vincentStartLogin`, opens the returned `authUrl`, and polls `/api/vincent/status`. The broken sessionStorage callback effect, client-side PKCE helpers, and `clearCodeVerifier()` on disconnect are gone.
- Existing poll-success and poll-timeout tests still pass against the new mock surface.

**Type safety cleanup** (`vincent-routes.ts`):

- Isolated the pre-existing `(config as any).vincent` casts behind typed `getVincentTokens` / `setVincentTokens` helpers. The remaining unsafe cast is now in one place (`as unknown as { vincent?: VincentTokens }`) with a comment explaining that `ElizaConfig` doesn't yet type the field.

## Flow after fix

1. User clicks "Login with Vincent" on the wallet page.
2. Desktop → `POST /api/vincent/start-login` → server returns `authUrl`.
3. Desktop opens `authUrl` in the external system browser via `desktopOpenExternal` RPC.
4. User authenticates on heyvincent.ai.
5. Vincent 302 → `http://127.0.0.1:<api-port>/callback/vincent?code=...&state=...`.
6. API handler validates `state`, exchanges code with Vincent (with matching `redirect_uri`), persists tokens, returns "You can close this window" HTML.
7. Desktop's 5-second status poll flips `vincentConnected` → `true`, UI shows success toast.

## Test plan

- [x] `bunx tsc -p packages/app-core/tsconfig.json --noEmit` — no new errors on Vincent files.
- [x] `bunx vitest run packages/app-core/src/state/useVincentState.test.ts` — 2/2 passing (existing poll-success and poll-timeout cases, updated mocks for the new client surface).
- [x] `bun run pre-review-local` via the milady `pre-review` agent — **APPROVE** (classification: `security`, code quality: pass, security: clear, tests: adequate, decision: APPROVE, exit 0).
- [x] Manual end-to-end smoke on `bun run dev:desktop`:
  - Wallet page → Login with Vincent → external browser opens
  - Auth on heyvincent.ai → redirected to `/callback/vincent` → "You can close this window" page
  - Desktop app flips to Connected; API log: `[vincent/callback] Vincent connected successfully`
  - Disconnect + reconnect round-trip verified: `[vincent/disconnect] Vincent disconnected` followed by a fresh `[vincent/callback] Vincent connected successfully`.

## Notes for reviewers

- The legacy `POST /api/vincent/register` and `POST /api/vincent/token` routes are still present but unused by the client. They can be deleted in a follow-up — left in place here to keep this PR scoped to the fix.
- `packages/app-core/src/api/vincent-oauth.ts` is now dead (client-side PKCE helpers) but left alone for the same reason.
- The desktop app caches dev bundles; reviewers validating locally should `kill` any stale `bun run dev` and restart so the new routes load.

### DIFF
--- a/packages/app-core/src/api/client-vincent.ts
+++ b/packages/app-core/src/api/client-vincent.ts
@@ -13,8 +13,15 @@ export interface VincentStatusResponse {
 
 // ── Declaration merging ───────────────────────────────────────────────
 
+export interface VincentStartLoginResponse {
+  authUrl: string;
+  state: string;
+  redirectUri: string;
+}
+
 declare module "./client-base" {
   interface MiladyClient {
+    vincentStartLogin(appName?: string): Promise<VincentStartLoginResponse>;
     vincentRegister(
       appName: string,
       redirectUris: string[],
@@ -30,6 +37,13 @@ declare module "./client-base" {
 }
 
 // ── Implementation ────────────────────────────────────────────────────
+
+MiladyClient.prototype.vincentStartLogin = async function (appName?: string) {
+  return this.fetch("/api/vincent/start-login", {
+    method: "POST",
+    body: JSON.stringify({ appName: appName ?? "Milady" }),
+  });
+};
 
 MiladyClient.prototype.vincentRegister = async function (
   appName: string,
--- a/packages/app-core/src/api/server.ts
+++ b/packages/app-core/src/api/server.ts
@@ -782,8 +782,20 @@ async function handleMiladyCompatRoute(
   }
 
   // ── Vincent OAuth routes ────────────────────────────────────────
-  if (url.pathname.startsWith("/api/vincent/")) {
-    if (!ensureCompatApiAuthorized(req, res)) return true;
+  // /callback/vincent is the OAuth redirect target and is hit by the user's
+  // external system browser — it has no compat API token, so it must bypass
+  // ensureCompatApiAuthorized. The PKCE code_verifier stored server-side
+  // (keyed by the OAuth state param) is what actually authorizes the
+  // token exchange.
+  if (
+    url.pathname.startsWith("/api/vincent/") ||
+    url.pathname === "/callback/vincent"
+  ) {
+    if (
+      url.pathname !== "/callback/vincent" &&
+      !ensureCompatApiAuthorized(req, res)
+    )
+      return true;
     const vincentConfig = loadElizaConfig();
     const handled = await handleVincentRoute(req, res, url.pathname, method, {
       config: vincentConfig,
--- a/packages/app-core/src/api/vincent-routes.test.ts
+++ b/packages/app-core/src/api/vincent-routes.test.ts
@@ -1,0 +1,479 @@
+/**
+ * Route-level tests for vincent-routes.ts — the server-side PKCE flow that
+ * backs Vincent OAuth login.  These are the security-critical gate: state
+ * validation, PKCE-backed token exchange, OAuth error reflection, and token
+ * persistence all live behind GET /callback/vincent.
+ *
+ * What's covered here (per PR #1685 review):
+ *   1. GET /callback/vincent with missing state       → 400 HTML
+ *   2. GET /callback/vincent with unknown state       → 400 HTML ("expired")
+ *   3. GET /callback/vincent with oauth `error` param → 400 HTML, value escaped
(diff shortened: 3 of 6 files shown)


Dexploarer:
Addressed the REQUEST CHANGES on test coverage.

**Added `packages/app-core/src/api/vincent-routes.test.ts`** — 10 route-level tests against the real `handleVincentRoute` (mocked `fetch` for Vincent upstream + mocked `saveElizaConfig`):

- GET /callback/vincent missing code → 400 HTML
- GET /callback/vincent missing `state` → 400 HTML
- GET /callback/vincent unknown `state` → 400 HTML ("login expired")
- GET /callback/vincent with OAuth `error` param → 400 HTML, value escaped (XSS guard)
- Upstream token endpoint failure → 502 HTML
- **Happy path:** POST `/api/vincent/start-login` → GET `/callback/vincent?code=…&state=…` → `saveElizaConfig` called with `{ accessToken, refreshToken, clientId, connectedAt }`, 200 HTML returned, and the token exchange body includes `redirect_uri` (RFC 6749 §4.1.3 regression guard)
- **State is single-use:** replaying the same state after a successful callback returns 400 and does not call `saveElizaConfig` again
- GET `/api/vincent/status` reflects persisted tokens and disconnected state
- POST `/api/vincent/disconnect` clears tokens via `saveElizaConfig`

**Bug found by the XSS test and fixed in the same commit:** the OAuth `error` query param was being double-escaped. The handler called `escapeHtml(oauthError)` and then passed the result into `sendCallbackHtml`, which already escapes `message` wholesale — so `<script>` became `&amp;lt;script&amp;gt;` in the rendered page (user saw literal `&lt;` text instead of rendered `<`). The defense was correct (no script ran) but the display was wrong. Removed the manual pre-escape; the test now asserts the output contains `&lt;script&gt;alert(1)&lt;/script&gt;` literally and does NOT contain the raw tag.

12/12 Vincent tests green locally (`vincent-routes.test.ts` + `useVincentState.test.ts`). Commit `81ff5a3b2`.

The unrelated E2E failure on `packages/agent/test/native-modules.e2e.test.ts > canvas can be imported` is pre-existing on develop and not touched by this PR.

PR #1709, opened 2026-04-06
fix(app): align Vite dep optimize target with es2022

dutchiono:
## Summary
- set `optimizeDeps.esbuildOptions.target` to `es2022` in app Vite config
- keep dep prebundle target aligned with existing app build/esbuild target
- add a regression test that asserts the app config keeps both targets aligned

## Why
Vite dependency optimization was running against the default legacy modules target (`chrome87`/`safari14` family), which caused esbuild transform failures on modern dependency code during desktop dev startup.

## Validation
- `bunx vite optimize --config vite.config.ts --force` (from `apps/app`)
- `bunx vitest run scripts/vite-app-target.test.ts`

### DIFF
--- a/.github/actions/setup-bun-workspace/action.yml
+++ b/.github/actions/setup-bun-workspace/action.yml
@@ -76,6 +76,10 @@ runs:
         key: bun-${{ runner.os }}-${{ hashFiles('bun.lock') }}
         restore-keys: bun-${{ runner.os }}-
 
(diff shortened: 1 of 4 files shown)


dutchiono:
Recycling PR event to trigger full CI on latest head SHA after workflow bootstrap fix.

dutchiono:
Superseded by #1710 from origin branch so full CI can run on the correct upstream ref.

PR #1710, opened 2026-04-06
fix(app): align Vite dep optimize target with es2022

dutchiono:
## Summary
- set optimizeDeps.esbuildOptions.target to s2022 in app Vite config
- keep dep prebundle target aligned with existing app build/esbuild target
- add a regression test that asserts the app config keeps both targets aligned
- initialize plugins/plugin-agent-orchestrator submodule in shared CI setup before un install
- apply formatter-only fixes in pps/app/plugins/mobile-signals/src/web.ts to satisfy lint baseline

## Why
- Fixes desktop dev Vite crashes where dep optimization was using legacy default target (chrome87/safari14) instead of project s2022.
- Fixes CI bootstrap failures where un install fails before tests due missing workspace path for plugins/plugin-agent-orchestrator.

## Validation
- unx vitest run scripts/vite-app-target.test.ts
- un run lint
- un run typecheck
- hook pre-review passed on push to origin


dutchiono:
Superseded by split PR #1712 for CI/runtime scope. Keeping Vite/UI fix isolated in #1708.

PR #1739, opened 2026-04-07
fix(app): clean up plugin restart follow-ups

dutchiono:
## Summary
- clear local plugin release-stream state even when uninstall restart does not come back online
- make the plugin restart test use a real AgentStatus shape

## Validation
- bunx @biomejs/biome check packages/app-core/src/components/pages/PluginsView.tsx packages/app-core/src/components/pages/PluginsView.install-restart.test.tsx
- bunx vitest run packages/app-core/src/components/pages/PluginsView.install-restart.test.tsx

### DIFF
--- a/apps/app/electrobun/src/__tests__/bridge-runtime.test.ts
+++ b/apps/app/electrobun/src/__tests__/bridge-runtime.test.ts
@@ -4,6 +4,8 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 const electroviewInstances = vi.fn();
 const defineRpc = vi.fn();
+const BOOT_CONFIG_STORE_KEY = Symbol.for("milady.app.boot-config");
+const BOOT_CONFIG_WINDOW_KEY = "__MILADY_APP_BOOT_CONFIG__";
 
 vi.mock("electrobun/view", () => {
   const Electroview = vi.fn(function MockElectroview(options: unknown) {
@@ -27,6 +29,8 @@ describe("electrobun bridge runtime", () => {
     delete w.__MILADY_API_BASE__;
     delete w.__MILADY_API_TOKEN__;
     delete w.__electrobun;
+    delete w[BOOT_CONFIG_WINDOW_KEY];
+    delete w[BOOT_CONFIG_STORE_KEY];
   });
 
   afterEach(() => {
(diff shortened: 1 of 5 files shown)


dutchiono:
1. Classification (derived): bug fix + test cleanup — consistent with author's claim. Three sub-changes: (a) uninstall-restart failure no longer leaks pluginReleaseStreams state, (b) test mocks switched to the real AgentStatus shape, (c) Windows smoke script assertions updated `$PID` -> `$currentPid` to match the actual `.ps1`.

2. Rubric (derived): UI state management correctness on error path, test mock fidelity, script assertion accuracy against the live PowerShell file.

3. Scope verdict: in scope

4. Universal invariants: all intact

5. Judgment:

Needed? — OK. The leak is real; the mock shape correction removes a suppressed type error; the assertion drift is a real CI regression.

Better than existing? — OK. Two call sites justify the `useCallback` extraction.
Duplication? — OK.
Blast radius covered? — CONCERN: `handleInstallPlugin` (line 571) and `handleUpdatePlugin` (line 648) have identical `if (!restarted) return;` guards that also skip the stream clear. Only the uninstall path was fixed. If the asymmetry is intentional (plugin still present on failed install/update, preserve stream for retry), that needs a comment.
Logic sound? — OK. Early-exit fix correct; empty dep array on `useCallback` correct (stable setter).
Complexity appropriate? — OK.
Tested meaningfully? — CONCERN: The actual production fix — uninstall failure path — has no test. The file only covers install. No test verifies that `pluginReleaseStreams` is cleared when uninstall's `restartAndWait` returns `mockAgentStatus("stopped")`.
Matches conventions? — OK.
Plausible breakage mode: Low risk. `pluginId` is a function argument, not a stale closure.

6. PR-type-specific checks:

Script assertion ↔ actual `.ps1`: OK — `smoke-test-windows.ps1` line 57 confirmed uses `$currentPid`
No UI code in `apps/app/src/` thin shell: OK
StartupPhase regression: N/A

7. Security: clear

8. Decision: REQUEST CHANGES

Two blockers:

- Add a test for the uninstall failure path — test that `restartAndWait` returning `mockAgentStatus("stopped")` during uninstall results in `pluginReleaseStreams` being cleared. Should fail without the fix.
- Document install/update intent — either fix the parallel `!restarted` returns in `handleInstallPlugin`/`handleUpdatePlugin` to also clear the stream, or add a comment explaining why leaving the stream intact there is deliberate.

PR #1756, opened 2026-04-08
fix: keep desktop cloud login on local route

dutchiono:
## Summary
- keep desktop cloud login on the local `/api/cloud/*` route even when no API base is persisted yet
- avoid the direct browser `cloudLoginDirect()` path in Electrobun, which is brittle under CORS in packaged shells
- add regression coverage for both decisions:
  - desktop chooses local route in `useCloudState`
  - `client.cloudLogin()` still resolves to same-origin `/api/cloud/login` when no base URL is persisted

## Included baseline gate fixes
- keep local maintainer gate (`pre-review:local`) green on current `develop` by carrying the same small baseline fixes already required in this checkout:
  - `scripts/run-biome-check.mjs`
  - `scripts/setup-upstreams.mjs`
  - `scripts/setup-upstreams.test.ts`
  - `packages/agent/package.json`
  - `packages/agent/src/external-modules.d.ts`
  - `packages/app-core/src/ambient-modules.d.ts`

## Validation
- `bun run pre-review:local`
- `bun run verify:lint`
- `bun run typecheck`
- `bunx vitest run packages/app-core/src/api/client.api-base.test.ts packages/app-core/src/state/useCloudState.test.ts`

## Notes
- this head does **not** include the previous `keyup` removal or trajectory contract churn
- cloud behavior change remains desktop-only; non-desktop/no-backend contexts still use direct auth

### DIFF
--- a/packages/agent/src/external-modules.d.ts
+++ b/packages/agent/src/external-modules.d.ts
@@ -1,10 +1,21 @@
 declare module "@elizaos/plugin-coding-agent";
 declare module "@elizaos/plugin-agent-orchestrator";
 declare module "@elizaos/plugin-agent-skills";
+declare module "@elizaos/plugin-anthropic";
+declare module "@elizaos/plugin-cron";
+declare module "@elizaos/plugin-experience";
 declare module "@elizaos/plugin-elizacloud";
+declare module "@elizaos/plugin-local-embedding";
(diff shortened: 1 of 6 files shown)


dutchiono:
Closing to restart clean from latest develop as requested. We will open a fresh minimal PR for the desktop cloud-login routing fix only.

PR #1770, opened 2026-04-09
fix: harden Windows repo setup

dutchiono:
## Summary
- treat plugin submodules as incomplete until their workspace package manifests exist
- repair empty initialized submodule worktrees before declaring setup success
- add Windows-safe plugin bin shims and a `bunx tsup` fallback during upstream setup
- rebuild plugin outputs when `dist/` exists but declared type artifacts are missing
- refresh `bun.lock` so frozen installs match the manifests on this branch

## Why
Windows installs could fail even after `git pull` because required plugin submodules were present but empty, and plugin build steps relied on bin-resolution behavior that was brittle under Bun on Windows.

CI also surfaced a separate lockfile problem on this branch: `bun install --frozen-lockfile --ignore-scripts` wanted to rewrite `bun.lock` under the same `MILADY_SKIP_LOCAL_UPSTREAMS=1` setup used by the Windows and test workflows. Refreshing the lockfile fixes that determinism issue.

## CI Notes
- `MILADY_SKIP_LOCAL_UPSTREAMS: "1"` is set in the Windows-focused workflows because CI should not depend on a repo-local `eliza` checkout.
- Windows frozen installs stay pinned to Bun `1.3.9` only where Bun `1.3.11` still reports false root lockfile drift; the broader matrix in `windows-dev-smoke.yml` remains on `1.3.10` and `canary` to keep coverage.
- The lockfile refresh also removes stale workspace lock entries that were no longer represented by the actual manifests visible to Bun on this branch.

## Review Notes
- The `bun run build` fallback in `scripts/setup-upstreams.mjs` is still present; one review appears to have relied on a truncated GitHub patch view that cut off before the unconditional fallback.
- `normalizeTsupVersionSpec` already falls back non-exact specs such as `workspace:^8` to `latest`, and that path is covered by tests.
- `SUBMODULE_READINESS_MARKERS` still take precedence over the generic plugin-manifest fallback.
- The `app-manager.test.ts` Hyperscape assertion updates are intentional test corrections for the runtime-backed launch path, not formatter-only churn.

## Validation
- `bun install --frozen-lockfile --ignore-scripts` with and without `MILADY_SKIP_LOCAL_UPSTREAMS=1`
- `bunx bun@1.3.9 install --frozen-lockfile --ignore-scripts` with and without `MILADY_SKIP_LOCAL_UPSTREAMS=1`
- `bunx vitest run scripts/init-submodules.test.ts packages/app-core/src/runtime/init-submodules-script.test.ts scripts/setup-upstreams.test.ts`
- `bun run typecheck`
- `bun run check` *(currently still reports the existing Biome docs-ignore issue in `apps/web/src/docs/content/**`, unrelated to this PR's changes)*

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -172,13 +172,7 @@ jobs:
           # Published-only CI restores the required non-eliza submodules explicitly.
           submodules: false
 
-      - name: Initialize Windows-safe workspace submodules
-        if: ${{ matrix.os == 'windows-latest' }}
-        shell: bash
-        run: git submodule update --init --depth 1 plugins/plugin-agent-orchestrator
-
       - name: Initialize tracked workspace submodules
-        if: ${{ matrix.os != 'windows-latest' }}
         run: node scripts/init-submodules.mjs
 
       - name: Setup Node.js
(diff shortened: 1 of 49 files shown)


dutchiono:
Consolidated into #1774. Closing this PR so the recovery work is reviewed and validated in one place.

PR #2016, opened 2026-04-21
fix(plugin-build): skip OS-locked native plugins on non-matching hosts

dutchiono:
## Summary
- skip native plugin builds when a plugin declares a pure OS host allowlist that excludes the current host
- support both `milady.platforms` and `elizaos.platforms` metadata in plugin package manifests
- keep mixed runtime target arrays (like `node`/`browser`) buildable on all hosts
- add unit tests for host-filter behavior in `apps/app/test/plugin-build-host-filter.test.ts`

## Why
Windows builds were failing when macOS-only native plugins were added without host guards.

## Validation
- `bunx @biomejs/biome check apps/app/scripts/plugin-build.mjs apps/app/test/plugin-build-host-filter.test.ts`
- `cd apps/app && bunx vitest run --config vitest.config.ts test/plugin-build-host-filter.test.ts`

> [!NOTE]
> ### Skip building native plugins on non-matching host OS in plugin-build script
> - Adds `shouldBuildPluginForHost()` in [plugin-build.mjs](https://github.com/milady-ai/milady/pull/2016/files#diff-5c979764d024bd7668278995923250da9ef1901a540cd9d9b9a47df12904e959) that reads `milady.platforms` or `elizaos.platforms` from each plugin's `package.json` and skips the build when the host OS is not in a pure OS allowlist.
> - Wraps build logic in an async `main()` and adds a direct-run guard so importing the module no longer triggers builds.
> - Adds a CI stub and override mappings for `@elizaos/plugin-app-control` so installs resolve to a local stub when the package is unpublished.
> - Adds a `git submodule sync --recursive` step in [init-submodules.mjs](https://github.com/milady-ai/milady/pull/2016/files#diff-927ba7fd8e1b3fbce8af2599149349017e91a9f97cab5d670b967df7f77ba209) before per-submodule initialization to keep remote URLs in sync with `.gitmodules`.
> - Risk: [audit-actions.mjs](https://github.com/milady-ai/milady/pull/2016/files#diff-2bbb6c2f3fba56ebbd0b6d91ecbac0656c425000b3024d0a2624b3b3c36d41a3) references `handlerReferencesText` after its declaration was removed, causing a `ReferenceError` at runtime when that code path is hit.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 11a53eb.</sup>
> 
>

PR #2017, opened 2026-04-21
fix(ci): map legacy wechat plugin import to workspace package

dutchiono:
## Summary
- add a compatibility Vite alias so `@miladyai/plugin-wechat` resolves to the local workspace plugin source (`eliza/plugins/plugin-wechat/src/index.ts`)
- add matching TypeScript path mappings in root `tsconfig.json` for `@miladyai/plugin-wechat`

## Why
Current `develop` CI fails in multiple workflows because `eliza/packages/app-core/src/config/wechat-config.ts` imports `@miladyai/plugin-wechat`, but the workspace package is `@elizaos/plugin-wechat`.

This patch avoids an immediate cross-repo submodule bump chain and unblocks Milady CI while upstream package naming converges.

## Validation
- `bunx @biomejs/biome check tsconfig.json apps/app/vite.config.ts`
- local full typecheck/build could not be run in this clean worktree because dependencies are not fully installed here (`vite/client` types missing), so CI is the source of truth for this fix.

> [!NOTE]
> ### Map legacy `@miladyai/plugin-wechat` import to workspace package source
> - Adds Vite alias and TypeScript path mappings pointing `@miladyai/plugin-wechat` to `eliza/plugins/plugin-wechat/src` so the browser build resolves the plugin from the local workspace instead of a published package.
> - Adds `@elizaos/plugin-app-control` to the CI stub overrides in [setup-upstreams.mjs](https://github.com/milady-ai/milady/pull/2017/files#diff-61d81dba031f233c46c15e4239a053eb2d3c374a926939e24aba9abbf8c7df91) and removes it from the root `package.json` dependencies.
> - Extends [restore-local-eliza-workspace.mjs](https://github.com/milady-ai/milady/pull/2017/files#diff-64f12dc4c1140282c6a043e6ee7414f35bb48fa1dcedff68851c99e70b602370) to reapply unpublished plugin stub overrides into `eliza/package.json` on every restore.
> - Adds TypeScript `ignoreDeprecations: "5.0"` and a lucide-react icon compatibility patch applied automatically during `setupUpstreams`.
> - Updates the mobile CI workflow to set an explicit iOS simulator destination/SDK and accept Android SDK licenses non-interactively.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 92fcb08.</sup>
> 
>


lalalune:
Merged the durable parts into develop as [`ed8afe321`](https://github.com/milady-ai/milady/commit/ed8afe321b) (+ [`06327f7e0`](https://github.com/milady-ai/milady/commit/06327f7e0a) to repair unintended test-file truncation).

**The headline fix is no longer needed.** `eliza/packages/app-core/src/config/wechat-config.ts` was reworked earlier today to inline the tiny package-name constant and `isWechatConfigured` shape-check. It no longer imports from `@miladyai/plugin-wechat` (or `@elizaos/plugin-wechat`) at all, so the Vite alias and tsconfig paths that this PR adds are obsolete.

**Kept from this PR**
- `.github/workflows/mobile-build-smoke.yml` — pin iOS destination/SDK env and auto-accept Android SDK licenses.
- `scripts/setup-upstreams.mjs` — `applyTypeScriptIgnoreDeprecationsCompatPatch` + `applyLifeOpsLucideCompatPatch` wired into both the skip-path and the real-install path; added `@elizaos/plugin-app-control` to `UNPUBLISHED_ELIZA_PLUGIN_CI_STUBS` with the correct `plugins/plugin-app-control/typescript` workspace entry.
- `scripts/restore-local-eliza-workspace.mjs` — reapply unpublished plugin stub overrides to `eliza/package.json` on every restore.
- `scripts/restore-local-eliza-workspace.test.ts` / `scripts/setup-upstreams.test.ts` — the new coverage for the above.

**Dropped (intentional)**
- `apps/app/tsconfig.json`, `apps/app/vite.config.ts`, `tsconfig.json` — `@miladyai/plugin-wechat` aliases are obsolete (see above).
- `apps/app/src/main.tsx` — same `register-slots` removal as #2016; would break the coding-agent UI slot wrappers. Also removes `applyLaunchConnection` usage in a way that silently drops its book-keeping in `handleDeepLink`.
- `.github/workflows/release-electrobun.yml` — the `ElizaOSApp-Setup-*.exe` → `Milady-Setup-*.exe` rename is only partial (4 of 9 occurrences) and the canonical Inno output is still `ElizaOSApp`, so it would break the Windows release workflow.
- `bun.lock` / `package.json` removal of `@elizaos/plugin-app-control` — it's already out of the root deps on develop, so that hunk is a no-op.

PR #2048, opened 2026-04-26
feat(patches): elizacloud bridge mirror — JSON output + fence-strip

2-A-M:
## Summary

Captures two patches that have been living in `node_modules/@elizaos/plugin-elizacloud` since alpha.4 in a downstream Milady checkout, and adds a postinstall step that re-applies them after every `bun install`.

The same two patches are being upstreamed in **[elizaos-plugins/plugin-elizacloud#15](https://github.com/elizaos-plugins/plugin-elizacloud/pull/15)**. Once that PR merges and a new alpha publishes, this bridge can be reverted in a single commit. Until then, this keeps Milady installs working out of the box.

## What the patches do

Both target `dist/{node,cjs}/index.node.{js,cjs}` in `@elizaos/plugin-elizacloud@2.0.0-alpha.8`:

1. **JSON output enforcement** — adds `text: { format: { type: "json_object" } }` to the `/responses` request body in `generateObjectByModelType`. Without this, the cloud plugin can ignore the caller's `schema` parameter and the model returns prose (`"I'll help you..."`) or markdown-fenced JSON, both of which choke `JSON.parse` downstream.

2. **Fence-strip before parse** — strips leading/trailing markdown code fences from the response text before `JSON.parse`. Models routinely wrap structured output in ` ```json ... ``` ` even when JSON is requested. The existing `getJsonRepairFunction()` doesn't handle leading backticks.

## What's added

- **`patches/elizacloud/0001-json-output-enforcement-and-fence-strip.patch`** — unified diff against `@elizaos/plugin-elizacloud@2.0.0-alpha.8` covering both hunks across both build outputs (4 hunks total).
- **`scripts/patch-elizacloud.mjs`** — idempotent applier: resolves the symlinked `node_modules` entry to its workspace path, version-pinned to alpha.8 (errors loudly on drift), reverse-checks first (no-op when already applied), then forward-applies.
- **`scripts/milady-postinstall-repo-setup.mjs`** — extended with a `miladyBridgePatchScripts` array. Runs after `runRepoSetup(repoRoot)` so elizaOS's pipeline (`patch-deps`, `link-external-plugins`, etc.) gets a chance to rewrite `node_modules` first; our patches win on top.

## Why a Milady-only entry point (not in `eliza/.../run-repo-setup.mjs`)

Adding the patch step to elizaOS's `repoSetupSteps` would force every other consumer of `@elizaos/app-core` to provide a `scripts/patch-elizacloud.mjs` in their own repo root, or installs would fail. Keeping it in Milady's own postinstall hook avoids that blast radius.

## Verified

- Reverse-apply strips both patches cleanly → 0 markers in dist files.
- Forward apply via script → 4 markers in dist files (Patch A + Patch B in each of 2 files).
- Re-run is idempotent → "patches already applied".
- Version-mismatch path errors with a clear message + non-zero exit.
- Round-trip: `rm -rf node_modules && bun install` → patches reapplied automatically; canonical Gmail→Discord workflow generation works on first try.

## Lifecycle

| Phase | Action |
|---|---|
| Now | Merge this PR; bridge keeps `bun install` survivable until upstream alpha publishes the fix. |
| When [elizaos-plugins/plugin-elizacloud#15](https://github.com/elizaos-plugins/plugin-elizacloud/pull/15) merges | Bump `@elizaos/plugin-elizacloud` dep to the new alpha. |
| Then | Delete `patches/elizacloud/`, `scripts/patch-elizacloud.mjs`, and the `miladyBridgePatchScripts` array entry. One commit. |

> [!NOTE]
> ### Add elizacloud bridge patch to enforce JSON output and strip markdown fences from model responses
> - Adds [0001-json-output-enforcement-and-fence-strip.patch](https://github.com/milady-ai/milady/pull/2048/files#diff-3aa53d7adb5dc89a1eafe61864ef88936a42299915bdbf5cfff28eb119098cf4) which modifies `generateObjectByModelType` in `@elizaos/plugin-elizacloud` dist artifacts to include `text: { format: { type: "json_object" } }` in request bodies and strip markdown code fences before `JSON.parse`.
> - Adds [patch-elizacloud.mjs](https://github.com/milady-ai/milady/pull/2048/files#diff-4917eb675fee6cbcddc73b6f937c9a67b107f6cf1dbad369720388dac3347e1b) which applies the patch only when the installed plugin matches the pinned version (`2.0.0-alpha.8`), skips if already applied, no-ops if the plugin is absent, and exits non-zero on version mismatch or patch failure.
> - Extends [milady-postinstall-repo-setup.mjs](https://github.com/milady-ai/milady/pull/2048/files#diff-f19495e066c61c71c1e3d74546e022726f4597109b995ed3a799e0179519c84a) to spawn `patch-elizacloud.mjs` (and any future patch scripts) after `runRepoSetup()`, failing postinstall if a script exits non-zero.
> - Risk: postinstall will now fail hard if the plugin is installed at the wrong version, blocking environment setup until the pinned version matches.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fd3dd56.</sup>
> 
>

### DIFF
--- a/patches/elizacloud/0001-json-output-enforcement-and-fence-strip.patch
+++ b/patches/elizacloud/0001-json-output-enforcement-and-fence-strip.patch
@@ -1,0 +1,72 @@
+diff --git a/dist/node/index.node.js b/dist/node/index.node.js
+index 4fe5094b9..79a0f4796 100644
+--- a/dist/node/index.node.js
++++ b/dist/node/index.node.js
+@@ -1327,7 +1327,12 @@ async function generateObjectByModelType(runtime, params, modelType, getModelFn)
+   const requestBody = {
+     model: modelName,
+     input,
+-    max_output_tokens: params.maxTokens ?? 8192
++    max_output_tokens: params.maxTokens ?? 8192,
++    // Milady patch: enforce JSON output at the API layer. Without this, the
++    // cloud plugin ignores the caller's `schema` parameter and the model
++    // returns prose ("I'll help you..." or markdown-fenced JSON), choking
++    // JSON.parse downstream.
++    text: { format: { type: "json_object" } }
+   };
+   if (!reasoning && typeof params.temperature === "number") {
+     requestBody.temperature = params.temperature;
+@@ -1366,10 +1371,17 @@ async function generateObjectByModelType(runtime, params, modelType, getModelFn)
+       totalTokens: data.usage.total_tokens ?? 0
+     });
+   }
+-  const jsonText = extractResponsesOutputText(data);
++  let jsonText = extractResponsesOutputText(data);
+   if (!jsonText.trim()) {
+     throw new Error("Object generation returned empty response");
+   }
++  // Milady patch: strip leading/trailing markdown code fences before JSON.parse.
++  // Models routinely wrap structured output in ```json ... ``` even when JSON
++  // is requested. The repair function does not handle the leading backtick.
++  jsonText = jsonText
++    .replace(/^[\s]*```(?:json)?\s*\n?/i, "")
++    .replace(/\n?```\s*$/i, "")
++    .trim();
+   try {
+     return JSON.parse(jsonText);
+   } catch (error) {
+diff --git a/dist/cjs/index.node.cjs b/dist/cjs/index.node.cjs
+index b496504b9..62772d917 100644
+--- a/dist/cjs/index.node.cjs
++++ b/dist/cjs/index.node.cjs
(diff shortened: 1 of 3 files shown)


lalalune (approved):
Reviewed the bridge patch and postinstall wiring. The change is scoped to the Milady postinstall path, version-pins the elizacloud artifact patch, is idempotent via reverse-check, and now passes local Biome check for the modified script. I pushed a formatting-only fix in fd3dd566.

PR #2090, opened 2026-05-03
fix(ci): windows release — disk cleanup, vitest/playwright crash, contract suite

NubsCarson:
## Summary
- adds disk cleanup steps before the electrobun desktop build and smoke test phases on windows CI runners — was running out of disk space every run
- switches smoke test to run against the build tree directly instead of extracting the 758MB installer into ~1.5GB (exceeds CI runner disk). installer is verified separately by the installer proof step
- **fixes vitest/playwright crash** — `live-provider.ts` has a top-level `import { test } from "vitest"` that loads `@vitest/expect` during playwright runs, crashing with `Cannot redefine property: Symbol($$jest-matchers-object)`. patches the import to lazy `await import("vitest")` at CI time before the playwright step
- removes 7 stale test file references from the release contract suite (files were deleted in 16f82e99 but refs left behind)
- fixes circleci config (wrong node version, npm instead of bun, missing postinstall)
- updates release-check compat patch for the removed `REQUIRE_INSTALLER` env var

supersedes #2089 — rebased clean onto latest develop. the setup-upstreams.mjs steward workspace check from that PR already landed.

**first successful windows release build in 20+ consecutive runs**: https://github.com/milady-ai/milady/actions/runs/25268152237/job/74086195443

## Test plan
- [x] PR CI checks pass (build, typecheck, lint, release contract)
- [x] full release-electrobun workflow — windows build passes end to end (57min, all green)
- [x] smoke test passes
- [x] playwright renderer bootstrap check passes
- [x] inno setup installer builds and verifies
- [x] MSIX package builds
- [ ] macOS builds (ARM failed due to apple timestamp service outage, not code — re-run when service is back)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2091, opened 2026-05-03
fix(release): cap changelog to 200 entries and truncate body at 100 KB

dutchiono:
## Summary

- GitHub's release API rejects bodies over 125,000 characters (the last run failed with a 257 KB body spanning ~2,945 commits between `v2.0.0-alpha.135` and the new tag)
- Two guards added to `agent-release.yml`:
  1. **Bash**: pipe `git log` through `head -200` so the file never grows past ~20 KB
  2. **JS**: truncate the `fullChangelog` string to 100,000 chars before embedding it in the release body, appending a note pointing users to the compare URL for the full history

## Test plan

- [ ] Re-run the Release workflow on `develop` — the `Tag & publish release` job should now create the release without hitting the 125 KB limit
- [ ] Verify the release body includes the compare URL so the full history is still reachable

> [!NOTE]
> ### Cap release changelog to 200 entries and truncate release body at 100 KB
> - Pipes the git commit list through `head -200` to limit changelog entries in both tagged and untagged release cases.
> - Truncates the generated release body to 100,000 characters, appending a truncation note when the limit is exceeded.
> - Also makes snap build continue on error, gates snap verify/upload steps on build success, and adjusts snapcraft staging path to `eliza/snap/snapcraft.yaml`.
> - Extends Debian packaging sed rewrites to cover `eliza.mjs` in addition to `elizaos-app.mjs`, and adjusts service path and `bun install` flags in `debian/rules`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0a4a587.</sup>
> 
>

### DIFF
--- a/.github/workflows/agent-release.yml
+++ b/.github/workflows/agent-release.yml
@@ -407,7 +407,6 @@ jobs:
       tag: ${{ needs.version.outputs.tag }}
       draft: false
       publish_release: false
-      channel: ${{ needs.version.outputs.channel }}
     secrets: inherit
 
   build-docker:
@@ -532,8 +531,8 @@ jobs:
           node scripts/disable-local-eliza-workspace.mjs
       - name: Stage snapcraft.yaml
         run: |
-          mkdir -p snap
-          cp eliza/packages/app-core/packaging/snap/snapcraft.yaml snap/snapcraft.yaml
+          mkdir -p eliza/snap
+          cp eliza/packages/app-core/packaging/snap/snapcraft.yaml eliza/snap/snapcraft.yaml
       - name: Normalize runner root ownership for snapd
         run: |
           set -euo pipefail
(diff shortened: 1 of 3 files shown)


lalalune (approved):
Changelog cap and release body truncation are correct and needed. Snap and debian CI fixes look solid. Dev scripts (dev-dutch.mjs / dev-dutch-pane.mjs) are fine as convenience tooling. One minor bug: dev-dutch-pane.mjs references undefined variable `l` (should be `plain`) in several regex checks in colorLine() — worth fixing in a follow-up but non-blocking since it's a dev-only script. Release Workflow Contract failure is a pre-existing issue that #2090 resolves. All other checks pass. LGTM.

lalalune:
All changes from this PR have already landed on develop directly (changelog cap at 72c5218a5e, snap/debian/dev-script fixes via subsequent commits). Closing as superseded.

PR #2101, opened 2026-05-04
fix(app): avoid privileged vite hmr bind

NubsCarson:
This PR will be reviewed by an AI agent. Provide clear context to help it assess your changes.

Aesthetic/UI changes that don't improve agent capability are out of scope and will be rejected.

## Category
- [x] Bug fix
- [ ] Security fix
- [ ] Performance improvement
- [ ] New feature
- [ ] Documentation
- [ ] Test coverage
- [ ] Other

## What

Separates Vite's local dev-server bind port from the public browser HMR client port.

## Why

When a public HTTPS origin is configured, the app should advertise the public `443` client port to browsers but still bind the local dev server to an unprivileged local port. Before this change, local runtime could try to bind privileged `:443` and fail with `EACCES`.

## How

Keeps the public-origin parsing for client HMR metadata, but resolves the actual dev-server bind port independently so local development does not attempt privileged binds.

## Testing

- Direct Bun assertion check against `apps/app/vite-dev-origin.ts` passed.
- Attempted direct Vitest path for `apps/app/test/vite-dev-origin.test.ts`; repo Vitest include config excludes that direct path, so the direct Bun assertion was used for focused verification.
- Live local Milady runtime no longer logs the privileged `:443` Vite bind failure after this change.

> [!NOTE]
> ### Fix Vite HMR config to bind on `uiPort` instead of the public origin port
> When `MILADY_VITE_ORIGIN`/`ELIZA_VITE_ORIGIN` is set, the HMR server was binding on the public origin's port, which can be a privileged port (e.g. 443). The fix sets `hmr.port` to `uiPort` for the local bind and introduces `clientPort` in [`vite-dev-origin.ts`](https://github.com/milady-ai/milady/pull/2101/files#diff-6164c3727b4562712e4e03edf46c76058258f312683319f188b66c750334bef4) to expose the correct public-facing port to HMR clients when it differs from `uiPort`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 88adc2a.</sup>
> 
>
>

### DIFF
--- a/apps/app/test/vite-dev-origin.test.ts
+++ b/apps/app/test/vite-dev-origin.test.ts
@@ -31,7 +31,8 @@ describe("resolveViteDevServerRuntime", () => {
       origin: "https://192.168.1.42",
       hmr: {
         host: "192.168.1.42",
-        port: 443,
+        port: 2138,
+        clientPort: 443,
         protocol: "wss",
       },
     });
@@ -50,7 +51,8 @@ describe("resolveViteDevServerRuntime", () => {
       origin: "http://10.0.0.2:9000",
       hmr: {
         host: "10.0.0.99",
-        port: 9000,
+        port: 2138,
+        clientPort: 9000,
         protocol: "ws",
       },
     });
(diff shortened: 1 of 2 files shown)


lalalune:
Integrated directly into develop in 6462a8e0dd. The current PR head is patch-equivalent to the develop change, and the shared eliza source/templates were also updated in milady-ai/eliza@eb4e889b83.

PR #2130, opened 2026-05-11
fix(patches/app-core): wire Capacitor.Plugins.Agent.start() + ChatView mobile-local bypass

NubsCarson:
## Summary

Four-hunk update to `patches/@elizaos%2Fapp-core@2.0.0-alpha.537.patch`, pairing with three upstream eliza commits on [`elizaOS/eliza#nubs/android-skip-vault-bootstrap-on-mobile-20260510`](https://github.com/elizaOS/eliza/tree/nubs/android-skip-vault-bootstrap-on-mobile-20260510).

Together, these unblock the stock-Android Capacitor APK so a fresh-install user can tap LOCAL → START LOCAL AGENT and land in a usable chat composer without manual adb tooling.

### Patch hunks

* **`RuntimeGate.js: resolveRuntimeChoices`** — unchanged from prior milady patch; mobile gets all three runtime tiles.
* **`RuntimeGate.js: client.setToken via ElizaNative`** — unchanged; AOSP/branded shells read the per-boot bearer through `window.ElizaNative.getLocalAgentToken()`.
* **`RuntimeGate.js: finishAsLocal`** (NEW) — alpha.537 set the runtime-mode pref + active-server but never dispatched `Capacitor.Plugins.Agent.start()`, so the foreground service never spawned and the WebView sat at "Initializing agent…" forever. Develop fixed this in `startMobileLocalAgent`; mirror inline. Also handles the bearer-token race (per-boot token is generated inside `ElizaAgentService.startAgentProcess()` on a background thread *after* the start() Intent dispatches — initial `client.setToken(...)` reads null and /api/status returns HTTP 401) with a 30s setInterval polling `ElizaNative.getLocalAgentToken()` that applies the bearer the moment the service writes it.
* **`ChatView.js`** (NEW) — alpha.537 locks the composer when `agentStatus.model === ""`. On mobile-local-runtime `/api/status` never reports a `model` field until the user picks a GGUF, so the composer stays locked indefinitely. Develop fixed this with a `readPersistedMobileRuntimeMode() === "local"` bypass; mirror by reading `localStorage["eliza:mobile-runtime-mode"]` directly.
* **`StartupShell.js` + `startup-phase-restore.js`** — unchanged.

## Test plan

* [x] Clean install on Moto G Play 2024 (Android 14, arm64) — picker → LOCAL → START LOCAL AGENT advances
* [x] bun spawns, /api/health returns 200 after ~30s model load, /api/status returns 200 with bearer auth
[...]

PR #2132, opened 2026-05-12
fix(apps/app): wire window.ElizaNative bearer prefetch + 90s watchdog into bootstrap

NubsCarson:
## Problem

On AOSP/Milady builds the agent runs in the same APK as the WebView and exposes its per-boot bearer through \`window.ElizaNative.getLocalAgentToken()\` (see \`os/android/.../ElizaNativeBridge.java\`). The current bootstrap-token chain in \`main.tsx\` only checks the URL fragment and \`localStorage\` — it never asks the native bridge. Result: on a fresh-install AOSP boot the React shell does not see the bearer and drops the user into \`PairingView\`, even though the device IS the agent.

There is a second window where the prefetch can't help: \`ElizaNativeBridge.getLocalAgentToken()\` returns \`null\` for the first ~30–50 s of app launch because the on-device agent process is still booting and hasn't written its bearer to its volatile static yet. The bootstrap is synchronous, so it fires before the bearer exists and would strand the shell in PairingView even with the prefetch wired in.

## Fix

Two changes, both in \`apps/app/src/main.tsx\`:

1. **Synchronous prefetch** — after the existing \`localStorage[SELF_HOSTED_TOKEN_KEY]\` read in the bootstrap chain, try \`window.ElizaNative?.getLocalAgentToken()\`. If we get a non-empty token, plant it back in \`localStorage\` so later reloads short-circuit the bridge.

2. **90s watchdog after \`setBootConfig\`** — if the bridge exists but the bearer isn't ready yet, poll every 500 ms for up to 90 s. First non-null token gets applied via \`client.setToken(...)\` + persisted to \`localStorage\`, then the loop exits.

No behavior change on stock Capacitor / iOS / Electrobun / web — \`window.ElizaNative\` is undefined there, so the synchronous block short-circuits and the watchdog returns before its first \`setTimeout\`.

## Verification

- Tested on Pixel 6a with the current milady debug APK: cold launch from a fresh install reaches the chat surface (Eliza greeting) without a PairingView round-trip. Watchdog applies the bearer ~35 s after launch when the on-device bun runtime first writes it.
- Stock Capacitor build (no ElizaNativeBridge): no measurable difference — \`getLocalAgentToken\` is undefined, both blocks no-op.
- Biome check: clean on the modified file.

## Relationship to #2130

This PR is the trimmed-down bearer-bootstrap piece from #2130, which got auto-closed by the agent-review workflow when both AI reviewers failed (API outage, not a code rejection). Refiling as a focused single-file PR so the review surface is just the bootstrap chain and the watchdog — no patches/scripts/skill churn.

> [!NOTE]
> ### Wire `window.ElizaNative` bearer token prefetch and 90s watchdog into app bootstrap
> - During bootstrap in [main.tsx](https://github.com/milady-ai/milady/pull/2132/files#diff-daadf90e7c659872f2d40ccafd9ee93306e0d519e96acb61016e83722e6f129a), if no token is found from the URL fragment or localStorage, the app now synchronously calls `window.ElizaNative.getLocalAgentToken()` and persists the result under `SELF_HOSTED_TOKEN_KEY` before applying the boot config.
> - If the synchronous prefetch yields no token, a 500ms polling watchdog runs for up to 90s, applying the native token to the client and localStorage as soon as it becomes available.
> - The watchdog exits immediately if a token is already set on the client, or if `window.ElizaNative` is unavailable.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b14ad34.</sup>
> 
>

### DIFF
--- a/apps/app/src/main.tsx
+++ b/apps/app/src/main.tsx
@@ -360,6 +360,27 @@ try {
       if (saved) bootstrapToken = saved;
     } catch {}
   }
+  // On AOSP/Milady the agent runs in the same APK and exposes its
+  // per-boot bearer through `window.ElizaNative.getLocalAgentToken()`
+  // (see `os/android/.../ElizaNativeBridge.java`). Pull it before the
+  // first auth-status poll so the React shell never hits PairingView
+  // for the loopback case where the device IS the agent.
+  if (!bootstrapToken) {
+    try {
+      const native = (
+        window as unknown as {
+          ElizaNative?: { getLocalAgentToken?: () => string | null };
+        }
+      ).ElizaNative;
+      const nativeToken = native?.getLocalAgentToken?.()?.trim();
+      if (nativeToken) {
+        bootstrapToken = nativeToken;
+        try {
+          window.localStorage.setItem(SELF_HOSTED_TOKEN_KEY, nativeToken);
+        } catch {}
+      }
+    } catch {}
+  }
   if (bootstrapToken) {
     appBootConfig.apiToken = bootstrapToken;
     appBootConfig.apiBase ??= window.location.origin;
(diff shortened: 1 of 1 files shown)


dutchiono (approved):
LGTM. Sync prefetch + 90s polling watchdog for AOSP bearer is sound. Guards on `bridge?.getLocalAgentToken` make it a no-op on non-AOSP builds. Uses recursive setTimeout (not setInterval) so polls don't overlap, bails on `client.hasToken()`, and stops cleanly at the 90s deadline. All CI green.

PR #2133, opened 2026-05-12
fix(patches): wire @elizaos/capacitor-llama + @elizaos/core npm patches into patchedDependencies

NubsCarson:
## Problem

Two patch files exist in milady's \`patches/\` directory on neighbour branches but are not wired into \`patchedDependencies\` in \`package.json\` on \`develop\`, so \`bun install\` doesn't apply them. The published-workflow build of the Capacitor APK (i.e. consuming \`@elizaos/*\` from npm, not \`MILADY_ELIZA_SOURCE=local\`) hits two known runtime failures at boot.

### Failure 1 — \`DEVICE_TIMEOUT\` from \`@elizaos/capacitor-llama@0.1.0\`

The published \`@elizaos/capacitor-llama@0.1.0\` stores the Capacitor proxy directly. \`await pluginLoadPromise\` calls \`proxy.then()\` on it, the Android Capacitor plugin has no \`then\` method, and the wait stays pending until the loader times out. Result: every mobile chat turn reports a generic "provider issue" and never reaches the on-device llama.cpp.

The upstream eliza source already has a \`toPlainLlamaCppPlugin\` wrapper that fixes this (and exposes the model's native Jinja \`getFormattedChat\` round-trip through dist + adapter). The published 0.1.0 predates that change. Milady's patch backports both fixes onto the installed dist.

This patch has a short half-life — the moment \`@elizaos/capacitor-llama\` publishes a version that incorporates the upstream changes from [elizaOS/eliza#7610](https://github.com/elizaOS/eliza/pull/7610) (now merged on \`elizaOS/eliza\` develop), the patch becomes a no-op and can be removed.

### Failure 2 — blank-screen WebView from \`@elizaos/core@2.0.0-alpha.537\`

\`@elizaos/core@2.0.0-alpha.537\`'s browser dist (\`dist/browser/index.browser.js\`) inlines fs-extra. fs-extra's top-level setup does:

\`\`\`js
typeof fs.realpath.native === "function"
  ? sI.realpath.native = ...
  : emitWarning(...);
\`\`\`

Fine on Node where \`fs.realpath\` is a function. In the Capacitor WebView with vite's \`node:fs\` stub, \`fs.realpath\` is sometimes undefined, the bare \`.native\` access throws \`Cannot read properties of undefined (reading 'native')\`, and the error happens at IIFE evaluation time *before* React mounts. The WebView shows a blank white screen and milady never boots past splash.

Milady's patch replaces the unsafe \`to.realpath.native\` access with a defensive guard on \`to.realpath\`, the assignment to \`sI.realpath.native\`, and the \`process.emitWarning\` call (process may not have \`emitWarning\` in the browser environment either). The fix falls through to fs-extra's intended fallback (warning emission) rather than crashing.

Longer half-life than the capacitor-llama patch — this fix is needed for as long as \`@elizaos/core\` inlines fs-extra at the top level of \`index.browser.js\`. Until that build choice changes upstream, every published-workflow milady mobile build needs this patch.

## Change

Two new patch files in \`patches/\`:

| File | Bytes | Target |
|---|---|---|
| \`patches/@elizaos%2Fcapacitor-llama@0.1.0.patch\` | 7,274 | toPlainLlamaCppPlugin wrapper + getFormattedChat wiring |
| \`patches/@elizaos%2Fcore@2.0.0-alpha.537.patch\` | 150,554 | fs.realpath.native guard in inlined fs-extra |

Two entries in \`package.json\` \`patchedDependencies\` map so \`bun install\` actually applies them.

## Scope

- ✅ Published-workflow Capacitor APK builds (the user-facing milady release path): both failures resolved at install time.
- ➖ \`MILADY_ELIZA_SOURCE=local\` workflow: unchanged — the source-link path bypasses the published \`@elizaos/*\` packages entirely, so the patches don't fire there.
- ➖ Desktop / Electrobun: unchanged — neither failure mode triggers off-mobile.

## Relationship to #2130

Originally bundled in #2130 alongside seven other commits (app-core patches, scripts, miladyos-dev.sh tweaks, etc.). That PR got auto-closed when both AI reviewers failed during a workflow API outage. Refiling as a focused patches-only PR so the review surface is just the two patch files + the two \`patchedDependencies\` lines.

> [!NOTE]
> ### Wire `@elizaos/capacitor-llama` and `@elizaos/core` patches into patchedDependencies
> - Registers two entries in [package.json](https://github.com/milady-ai/milady/pull/2133/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) `patchedDependencies` so the package manager applies local patches for `@elizaos/capacitor-llama@0.1.0` and `@elizaos/core@2.0.0-alpha.537` on install.
> - The capacitor-llama patch adds a `toPlainLlamaCppPlugin` wrapper to prevent the Capacitor plugin proxy from being treated as a thenable, fixing stalled load/generate operations on Android.
> - It also adds a `formatChat` method to `CapacitorLlamaAdapter` and a `formatChat` message handler in `DeviceBridgeClient` for native Jinja chat template formatting over WebSocket.
> - The core patch modifies the distributed browser bundle at `dist/browser/index.browser.js`.
> - Risk: patches are applied against specific version pins; version bumps to either package will silently skip the patches until entries are updated.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized be1e5ad.</sup>
> 
>
>


dutchiono:
The Lint & Format failure here isn't from this PR's changes — it's pre-existing biome formatter drift in `apps/app/vite.config.ts` on develop (introduced by the recent vite work in 855d96c83 / 19b3a7b1e / 3e4750528). I opened #2137 to fix develop. Once that lands, please rebase and lint should pass.

Content review: patches look right. `toPlainLlamaCppPlugin` wrapper correctly handles the Capacitor proxy thenable issue, and the `fs.realpath.native` guard is the right defensive fix for the inlined fs-extra in the browser bundle. Both patches are tightly version-pinned which the description acknowledges. Good explanation of half-life.

dutchiono:
#2137 merged — develop lint is unblocked. Please rebase this branch on develop and CI should go green.

dutchiono (approved):
Approved. `toPlainLlamaCppPlugin` wrapper is the right fix for the Capacitor proxy thenable issue on Android, and the `fs.realpath.native` guard is the right defensive fix for the inlined fs-extra in the browser bundle. Version-pinned patches with documented half-life. All CI green after rebase on develop.

PR #2151, opened 2026-05-14
fix(app): repair local-mode mobile vite build

NubsCarson:
## Summary

Repairs the local-source mobile app Vite build used by `MILADY_ELIZA_SOURCE=local` Android/APK work. The PR keeps the renderer on browser-safe stubs for server/native-only modules while still resolving the local Eliza workspace packages that the mobile build needs.

This branch has been merged with current `origin/develop` and the conflict in `apps/app/vite.config.ts` was resolved to keep both the newer package-mode alias work and the local-mode Android fixes.

## What changed

- Restores missing local app-core/shared helper resolution used by the local Eliza workspace path.
- Fixes scoped package detection in native-module stubbing so `@elizaos/*` packages match their configured stub generators instead of collapsing to `@elizaos`.
- Expands the `@elizaos/plugin-local-inference` stub surface for imports from the package root plus `/routes`, `/runtime`, and `/services`, including `handleLocalInferenceTtsRoute`.
- Makes the node-builtin function proxy invariant-safe by forwarding `ownKeys` and `getOwnPropertyDescriptor` through `Reflect`, avoiding the blank-WebView failure where React never mounted.
- Preserves upstream package-mode app-core/UI aliasing and CI/export stabilization from latest `develop`.

## Root cause

The local-source Android build resolves a much larger local Eliza graph than package-mode desktop/web builds. Server-only packages and native-only packages are still statically visible to Rollup even though the renderer never executes them. Several of those imports either lacked named-export stubs or hit a proxy invariant error during renderer initialization, which caused build failures or a blank WebView.

## Impact

The changes are scoped to app build-time aliasing/stubbing. Package-mode behavior is preserved, and the local-mode Android renderer can build without dragging native/server code into the WebView.

## Validation

- `bun test apps/app/test/package-mode-aliases.test.ts` — 7 passed.
- `MILADY_ELIZA_SOURCE=local ELIZA_CAPACITOR_BUILD_TARGET=android bun run --cwd apps/app build` — passed, 7504 Vite modules transformed.
- Full APK integration was also validated in the Pixel workspace using the same local-mode Vite surface: the APK booted past the previous white screen into the chat UI, and the local agent reached `/api/health` ready with 10 plugins loaded and 0 failed.

## Notes

This PR intentionally stays wrapper-side. The separate Eliza/native-runtime fixes are being split into their own PRs so maintainers can review the package-boundary, Android runtime, and local-inference changes independently.

### DIFF
--- a/apps/app/vite.config.ts
+++ b/apps/app/vite.config.ts
@@ -414,7 +414,22 @@ function resolveLocalSharedAliases(): Alias[] {
   return aliases;
 }
 
+/**
+ * When a local eliza workspace is present and `@elizaos/shared` has been
+ * built, return the path to its local `dist/index.js` so the renderer
+ * prefers it over the (possibly lagging) bun-store published copy.
+ */
+function resolveLocalSharedDist(): string | null {
+  const localDist = path.join(localElizaRoot, "packages/shared/dist/index.js");
+  return fs.existsSync(localDist) ? localDist : null;
+}
+
 function resolveLocalAppCoreAliases(): Alias[] {
+  const sharedDist = resolveLocalSharedDist();
+  const sharedDistDir = sharedDist
+    ? path.join(localElizaRoot, "packages/shared/dist")
+    : null;
+
   // Map @elizaos/agent's root import to its real source so static named
   // imports from compiled app-core code (e.g. account-pool.js's
   // ACCOUNT_CREDENTIAL_PROVIDER_IDS) resolve. Server-only runtime branches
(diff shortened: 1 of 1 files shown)


NubsCarson:
Closing as superseded by #2153.

#2151 only touched `apps/app/vite.config.ts` and is now conflict-dirty against `develop`. The Android/local-runtime package-mode cleanup has been carried forward into #2153 with broader validation: package-mode build, typecheck, lint, focused package-mode tests, and GitHub Actions build/type/lint/unit/release-contract passing. Keeping this older PR open would create duplicate review surface.

PR #2153, opened 2026-05-16
fix(app): stabilize Android local runtime boot

NubsCarson:
## Summary

Stabilizes the wrapper-side Android/local-runtime build path so Milady can build against the package-mode CI environment and the local Eliza runtime path without stale package drift.

This PR currently covers five verified areas:

- Android local-runtime wrapper polish: the WebView no longer tries to own embedded agent startup twice; Java `ElizaAgentService`/Agent transport remains the owner of local runtime startup and token injection.
- Package-mode UI cleanup: replaces broad legacy `elizaos-ui-*compat` shims with focused, app-owned voice capture / voice pill fallbacks, while keeping real package exports for `@elizaos/ui/browser`, style files, and source subpaths when available.
- Local/package source alignment: Vite and TypeScript resolve current `@elizaos/ui` browser/voice source paths, native plugin packages, and generated package layouts instead of stale `dist`/npm edges.
- CI/package-mode hardening: package-mode keeps package tsconfig paths when `eliza/` is absent, skips local Eliza runtime release-contract steps when the nested checkout is intentionally absent, and tsdown falls back through current bundled `@elizaos/app-core` layouts without relying on unexported `package.json` subpaths.
- Release/build cleanup: cssstyle/jsdom gets a reproducible CJS color alias patch, the static asset manifest is current, and the Electrobun release workflow carries the required Whisper model artifact prepare/upload/download/seed path.

The latest refresh also refactors the CodeFactor-reported complexity hotspots in `apps/app/vite.config.ts` into named helper functions and removes the `unicorn/no-useless-fallback-in-spread` issue from `scripts/pixel-local-voice-benchmark.mjs`.

Related Eliza core/native PR: elizaOS/eliza#7792.

## Current Source Base

Updated 2026-05-19 with normal merges, not rebases.

- PR branch head: `85335e9cb`
- Latest merged `origin/develop`: `8c10e9fb0`
- Previous source-refresh head: `6c8f89faa`
- CodeFactor cleanup commit: `175334bc4`
- This package-mode validation worktree intentionally has no nested `eliza/` checkout; local-Eliza-only release-contract steps are skipped by design in that mode.

## Validation

Local validation on `fbc44b75d` / `175334bc4` after the CodeFactor cleanup:

```bash
bun install --frozen-lockfile
bunx vitest run --passWithNoTests \
  apps/app/test/package-mode-aliases.test.ts \
  scripts/standalone-eliza-package-contract.test.ts \
[...]

PR #2157, opened 2026-05-19
fix(ci): guard auth gate package-mode builds

NubsCarson:
## Summary

Adds a small guard to the agent-review Auth P0 workflow so package-mode PR checkouts skip the local Eliza runtime plugin build when the nested `eliza/packages/core` checkout is absent.

## Why

Several package-mode Milady PRs can fail before their actual auth tests because the `pull_request_target` workflow assumes `eliza/packages/core` exists and then runs:

```sh
(cd eliza/packages/core && bun run build)
```

That directory is only present in local-source/nested-Eliza checkouts. In package-mode checkouts the workflow should skip this optional local runtime plugin prebuild instead of failing with `cd: eliza/packages/core: No such file or directory`.

This is intentionally split from the larger Android APK/runtime PR so maintainers can land the base workflow fix independently.

## Validation

```sh
git diff --check
node scripts/validate-ci-bootstrap-contract.mjs
```

Both passed locally.

## Notes

Because this workflow runs under `pull_request_target`, GitHub evaluates the workflow file from the base branch. This PR may not be able to make its own Auth P0 run observe the guard until the change lands on `develop`; the value here is unblocking subsequent PR checks once merged.

> [!NOTE]
> ### Skip local eliza runtime plugin build when `eliza/packages/core` is absent
> Adds a directory existence check at the start of the 'Build local eliza runtime plugins' step in [agent-review.yml](.github/workflows/agent-review.yml). If `eliza/packages/core` is missing, the step prints a message and exits successfully instead of running build commands that would fail.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5f777b4.</sup>
> 
>

### DIFF
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -155,6 +155,10 @@ jobs:
 
       - name: Build local eliza runtime plugins
         run: |
+          if [ ! -d eliza/packages/core ]; then
+            echo "eliza checkout is absent; skipping local runtime plugin build"
+            exit 0
+          fi
           if [ ! -f eliza/packages/core/dist/index.node.js ] || [ ! -f eliza/packages/core/dist/index.d.ts ]; then
             (cd eliza/packages/core && bun run build)
           fi


NubsCarson:
Closing this standalone guard PR to avoid PR noise. The guard itself is still valid, but this branch cannot independently pass current package-mode CI from bare develop because the UI/package-mode fixes in #2153 are also required (typecheck fails on missing @elizaos/ui exports before this guard matters). The same Auth P0 guard remains included in #2153, where it is paired with the package-mode fixes that make the branch buildable.

PR #2177, opened 2026-05-30
fix(app/dev): exclude @elizaos/agent runtime from renderer dep-prebundle

Dexploarer:
## Problem
`bun run dev` never brought up the renderer (blank/broken UI) and stalled: vite's esbuild `optimizeDeps` dep-prebundle scanned `@elizaos/agent` against the **published alpha** (which lags develop and is missing newer subpaths like `runtime/plugin-collector`) and crashed with `Missing "./runtime/plugin-collector" specifier`, retrying 6× then giving up.

## Fix
`@elizaos/agent` is the **agent runtime — it runs in its own process** (the `eliza-api` server on :31337); the renderer talks to it over HTTP and already **stubs** it via `nativeModuleStubPlugin`. The only bug was the dep-prebundle scan hitting the stale package before the stub applied. Add `@elizaos/agent` to `optimizeDeps.exclude` so the scan skips it and the stub handles it at build time — the runtime is **not** bundled into the renderer.

## Verified
`bun run dev`: vite now serves on **:2138 with no agent crash**; the agent runtime still boots in its own **:31337** process.

> [!NOTE]
> ### Exclude `@elizaos/agent` from Vite renderer dependency pre-bundling
> Adds `@elizaos/agent` to `optimizeDeps.exclude` in [vite.config.ts](https://github.com/milady-ai/milady/pull/2177/files#diff-0d80c2713ea1483c55bbc803b5d5d6a1f75654501e99a33c5d306cdbddd511e4) to prevent Vite from attempting to pre-bundle the agent runtime package in the renderer process.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 60f1be7.</sup>
> 
>

### DIFF
--- a/apps/app/vite.config.ts
+++ b/apps/app/vite.config.ts
@@ -2900,6 +2900,12 @@ export default defineConfig({
       // Contains native-only pty-state-capture / pty-console imports; skip pre-bundling.
       "@elizaos/plugin-agent-orchestrator",
       "pty-console",
+      // @elizaos/agent is server-side and the published alpha lags develop
+      // (missing newer subpaths like runtime/plugin-collector). esbuild
+      // dep-prebundling resolves it from the stale published package and
+      // crashes ("Missing ./runtime/plugin-collector specifier"); excluding it
+      // lets resolve.alias map @elizaos/agent[/*] to local source instead.
+      "@elizaos/agent",
       // Built-in secrets live in @elizaos/core features; Vite must not externalize them as a separate package.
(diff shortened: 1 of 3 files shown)


lalalune:
Superseded — the fix (`@elizaos/agent` in `optimizeDeps.exclude`) already landed on `develop` (`apps/app/vite.config.ts`, with the same "published alpha lags develop" rationale). This branch is 33 commits behind develop and its red CI is staleness, not a real signal. Closing as already-resolved.

